PR #295, opened 2015-01-01
Improving time_length for 29 Feb (when dealing with date time)

larmarange:
With previous version, we had:

```r
> naiss <- ymd_hms("1992-02-29 12:00:00")
> evt <- ymd_hms("2014-03-01 01:00:00")
> time_length(interval(naiss, evt), "years")
[1] 21.99874
```

We should consider that the anniversary occurs between 28 Feb at 23:59:59 and 1st March at 00:00:00. Therefore, `%m++%` should rollback at 00:00:00. In that case:

```r
> time_length(interval(naiss, evt), "years")
[1] 22.00011
```

### DIFF
--- a/R/intervals.r
+++ b/R/intervals.r
@@ -632,15 +632,12 @@ setMethod("time_length", signature("Interval"), function(x, unit = "second") {
     periods <- as.period(x, unit = unit)
     int_part <- slot(periods, unit)
 
-    neg <- x@.Data < 0
-    pos <- !neg
-    prev_aniv <- next_aniv <- int_start(x)
-
-    prev_aniv[pos] <- int_start(x)[pos] %m++% (int_part [pos] * period(1, units = unit))
-    next_aniv[pos] <- int_start(x)[pos] %m++% ((int_part[pos] + 1L) * period(1, units = unit))
-
-    prev_aniv[neg] <- int_start(x)[neg] %m+% (int_part[neg] * period(1, units = unit))
-    next_aniv[neg] <- int_start(x)[neg] %m+% ((int_part[neg] - 1L) * period(1, units = unit))
+    prev_aniv <- .month_plus(
+      int_start(x), (int_part * period(1, units = unit)), 
+      roll_to_first = TRUE, preserve_hms = FALSE)
+    next_aniv <- .month_plus(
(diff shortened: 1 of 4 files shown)


larmarange:
A quick comment about period computation from intervals:

```r
> naiss <- ymd_hms("1992-02-29 12:00:00")
> evt <- ymd_hms("2014-03-01 01:00:00")
> as.period(interval(naiss, evt))
[1] "22y 0m -1d 13H 0M 0S"
```

Should we avoid negative days?

vspinu:
This should not happen. It's a bug in as.period.

vspinu:
>> naiss <- ymd_hms("1992-02-29 12:00:00")
>> evt <- ymd_hms("2014-03-01 01:00:00")
>> time_length(interval(naiss, evt), "years")
> [1] 21.99874

Well. There is a mirror problem with negative intervals that use `%m+%`:

```
interval(ymd_hms('2000-02-29 12:00:00'), ymd_hms('1999-02-28 20:00:00'))
```

So we will have to change %m+% as well, and I am not sure that's a good idea. %m+% will basically make all intraday timestamps for Feb 29 equal, and that's clearly not suitable for intraday analysis. 

One easy solution is to add an optional argument to `rollback` and `.month_plus` and `.month_plus_plus`. Something like `preserve_hms = TRUE`. 

BTW. your `.rollback_day_one` can also be integrated into `rallback` with an optional argument  `roll_to_first = FALSE` or something similar.

larmarange:
You are right.

We can also add `roll_to_first` to `.month_plus`.   In that scenario, we don't need `%m++` and `.month_plus_plus`  (as it's not exported) and we can use directly `.month_plus` in `time_length`.

I will update the pull request.

larmarange:
When we think with time, even for negative intervals we should not use `%m-%`. In fact, the anniversary should be the 28 Feb at 24:00:00, i.e. the 1st March at 00:00:00.

Therefore, the code is simpler.

Now, we obtain:

```r
> time_length(interval(ymd_hms('2000-02-29 12:00:00'), ymd_hms('1999-02-28 20:00:00')), "years")
[1] -1.000457
```

larmarange:
Tests and examples added for rollback. Code of time_length updated with `ifelse(x@.Data < 0, -1, 1)`

vspinu:
Great! Thanks.

PR #302, opened 2015-01-30
Allow to parse dates from long numbers

jiho:
With as.character(), long numbers get converted to scientific
notation which the parse of course fails to recognise. Using
format(..., scientific = FALSE) prevents that.

### DIFF
--- a/inst/tests/test-parsers.R
+++ b/inst/tests/test-parsers.R
@@ -156,6 +156,16 @@ test_that("ymd functions correctly parse dates with no separators and no quotes"
               equals(as.POSIXct("2010-01-02 23:59:59", tz = "UTC")))
 })
 
+test_that("all numbers are correctly converted into character for parsing", {
+  # Conversion to character of numbers with 000000 at the end with as.character()
+  # may outputs scientific notation depending on the value of the 'scipen' option.
+  # .num_to_date() uses format() to avoid that. Check that it does indeed work
+  expect_that(ymd_hms(20100102000000),
+              equals(as.POSIXct("2010-01-02 00:00:00", tz = "UTC")))
+  expect_that(.num_to_date(20100102000000),
+              equals("20100102000000"))
+})
+
 test_that("ymd functions parse absurd formats as NA's", {
   ## should not through errors, just as as.POSIX and strptime
   pna <- as.POSIXct(as.POSIXlt(NA, tz = "UTC"))
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,8 @@
+
+BUG FIXES
+
+* allow to parse long numbers such as 20140911000000 as date+time (@jiho, #302)
+
 Version 1.3.3
 -------------------------------------------------------------------------
 
--- a/R/parse.r
+++ b/R/parse.r
@@ -695,7 +695,7 @@ fast_strptime <- function(x, format, tz = "UTC"){
   if (is.numeric(x)) {
     out <- rep.int(as.character(NA), length(x))
     nnas <- !is.na(x)
-    x <- as.character(x[nnas])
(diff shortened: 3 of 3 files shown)


jiho:
Should really have run `check()`. Additional commits fix mistakes.

jiho:
Example of what used to fail:

```
> x <- 20140101000000
> as.character(x)
[1] "2.0140101e+13"
> format(x, scientific = FALSE, trim = TRUE)
[1] "20140101000000"
> .num_to_date(x)
[1] "20140101000000"
```

It seems like a bug in `as.character` actually but it is probably safer to fix it here rather than wait for `as.character` to change...

vspinu:
Thanks. Could you please add a small test for this?

Could you also squash your commits into one? We try to stick to one-commit-per-change rule to make it easier to navigate the history.

jiho:
Regarding performance, `format` would actually be even a bit better:

```
> x <- round(runif(10^7) * 100000)
> system.time(as.character(x))
   user  system elapsed 
  7.786   0.015   7.816 
> system.time(format(x, scientific = FALSE, trim = TRUE))
   user  system elapsed 
  6.412   0.013   6.431 
```

jiho:
Tests are added and everything is squashed into one commit

inst/tests/test-parsers.R
```diff
@@ -156,6 +156,20 @@ test_that("ymd functions correctly parse dates with no separators and no quotes"
               equals(as.POSIXct("2010-01-02 23:59:59", tz = "UTC")))
 })
 
+test_that("numbers are correctly converted into character for parsing", {
+  # numbers with 000000 at the end are wrongly converted by as.character
```
vspinu: They are not wrongly converted. That's the feature which is customized by `scipen` option.

inst/tests/test-parsers.R
```diff
@@ -156,6 +156,20 @@ test_that("ymd functions correctly parse dates with no separators and no quotes"
...
+  # which outputs scientific notation
+  expect_that(ymd_hms(20100102000000),
+              equals(as.POSIXct("2010-01-02 00:00:00", tz = "UTC")))
+  expect_that(.num_to_date(20100102000000),
+              equals("20100102000000"))
+  # check for a fix in as.character
```
vspinu: I don't think we need these tests.  `as.character` is not broken and will never be fixed.

vspinu:
This PR made me think, are there other places `as.character` will fail similarly?

jiho:
Potential problems:

- `R/parse.r` l. 455 `as.character` is used with randomly generated numbers which may end with many zeros by chance.
- `inst/tests/test-parser.r` l. 520 same as above.
- `R/parse.r` l. 489, `as.character` is used when it is not necessary (`.num_to_date` always returns a `character`)

The other uses in the source code do not seem problematic to me.

jiho:
I've addressed the comments above. Let me know what you think. I guess I'll need to squash the commits again but I'll rather do that when all is accepted.

vspinu:
> R/parse.r l. 455 as.character is used with randomly generated numbers which may end with many zeros by chance.

Probably not an issue as `as.character` is applied to POSIXct object. `as.character(.POSIXct(1000000000000))` works as expected.

> R/parse.r l. 489, as.character is used when it is not necessary (.num_to_date always returns a character)

Good catch. That makes me think that `.num_to_date` is a misnomer. It doesn't convert to date.

> I've addressed the comments above. Let me know what you think. I guess I'll need to squash the commits again but I'll rather do that when all is accepted.

Yes. Looks good. Please squash. Instead of adding new commits you can always amend previous commit and then force push.

Thanks.

jiho:
Squashed commits

vspinu:
:+1: Thanks!

jiho:
You are welcome. Happy to help.

PR #303, opened 2015-02-05
Adding "quarter" option to floor/ceiling_date

jonboiser:
In reference to [Issue 239](https://github.com/hadley/lubridate/issues/239).

### DIFF
--- a/R/round.r
+++ b/R/round.r
@@ -2,14 +2,14 @@
 #'
 #' \code{floor_date} takes a date-time object and rounds it down to the nearest integer 
 #' value of the specified time unit. Users can specify whether to round down to 
-#' the nearest second, minute, hour, day, week, month, or year.
+#' the nearest second, minute, hour, day, week, month, quarter, or year.
 #'
 #' By convention the boundary for a month is the first second of the month. Thus
 #' \code{floor_date(ymd("2000-03-01"), "month")} gives "2000-03-01 UTC".
 #' @export floor_date
 #' @param x a vector of date-time objects 
 #' @param unit a character string specifying the time unit to be rounded to. Should be one of 
-#'   "second","minute","hour","day", "week", "month", or "year."
+#'   "second", "minute", "hour", "day", "week", "month", "quarter", or "year."
 #' @return x with the appropriate units floored
 #' @seealso \code{\link{ceiling_date}}, \code{\link{round_date}}
 #' @keywords manip chron
@@ -27,9 +27,11 @@
 #' # "2009-08-02 CDT"
 #' floor_date(x, "month")
 #' # "2009-08-01 CDT"
+#' floor_date(x, "quarter")
+#' # "2009-07-01 CDT"
 #' floor_date(x, "year")
 #' # "2009-01-01 CST"
-floor_date <- function(x, unit = c("second","minute","hour","day", "week", "month", "year")) {
+floor_date <- function(x, unit = c("second", "minute", "hour", "day", "week", "month", "year", "quarter")) {
   unit <- match.arg(unit)
   
   new <- switch(unit,
@@ -39,6 +41,7 @@ floor_date <- function(x, unit = c("second","minute","hour","day", "week", "mont
     day =    update(x, hours = 0, minutes = 0, seconds = 0),
     week =   update(x, wdays = 1, hours = 0, minutes = 0, seconds = 0),
     month =  update(x, mdays = 1, hours = 0, minutes = 0, seconds = 0),
+    quarter = update(x, months = ((month(x)-1)%/%3)*3+1, mdays = 1, hours = 0, minutes = 0, seconds = 0),
     year =   update(x, ydays = 1, hours = 0, minutes = 0, seconds = 0)
   )
   new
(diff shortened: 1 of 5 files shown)


R/round.r
```diff
@@ -76,25 +77,30 @@ floor_date <- function(x, unit = c("second","minute","hour","day", "week", "mont
...
+
 	switch(unit,
 		minute = minute(y) <- minute(y) + 1,
 		hour =   hour(y) <- hour(y) + 1,
 		day =    yday(y) <- yday(y) + 1,
 		week =   week(y) <- week(y) + 1,
 		month =  month(y) <- month(y) + 1,
-		year =   year(y) <- year(y) + 1
+		year =   year(y) <- year(y) + 1,
+        quarter = { month(y) <- month(y) + 3; y <- y - ddays(1) }
```
vspinu: Why do you need to subtract `ddays(1)`?

jonboiser: The issue reporter wanted ceiling to return the last day of the quarter. This line adds three months to the first day of the quarter, then subtracts one (to handle months with 28/30/31 days), instead of adding two months, and having to figure out how long a month is to get the last day.

vspinu: This introduces an exception for quarter and we don't want that. OP can substract 1 day himself if so desired.

R/round.r
```diff
@@ -76,25 +77,30 @@ floor_date <- function(x, unit = c("second","minute","hour","day", "week", "mont
...
+
 	switch(unit,
 		minute = minute(y) <- minute(y) + 1,
 		hour =   hour(y) <- hour(y) + 1,
 		day =    yday(y) <- yday(y) + 1,
 		week =   week(y) <- week(y) + 1,
 		month =  month(y) <- month(y) + 1,
-		year =   year(y) <- year(y) + 1
+		year =   year(y) <- year(y) + 1,
+        	quarter = month(y) <- month(y) + 3
```
jonboiser: Removed that `dday` from this line. The behavior of ceiling is now to return the first day of the next quarter. The behavior of floor is to return the first day of the nearest quarter.

R/round.r
```diff
@@ -39,7 +39,8 @@ floor_date <- function(x, unit = c("second","minute","hour","day", "week", "mont
     day =    update(x, hours = 0, minutes = 0, seconds = 0),
     week =   update(x, wdays = 1, hours = 0, minutes = 0, seconds = 0),
     month =  update(x, mdays = 1, hours = 0, minutes = 0, seconds = 0),
-    year =   update(x, ydays = 1, hours = 0, minutes = 0, seconds = 0)
+    year =   update(x, ydays = 1, hours = 0, minutes = 0, seconds = 0),
+    quarter = update(x, months = ((month(x) - 1) %/% 3)*3 + 1, mdays = 1, hours = 0, minutes = 0, seconds = 0)
```
jonboiser: Truncated the smaller time units.

R/round.r
vspinu: > Actually, what would happen for dates in Q4 (e.g. november)? 

What's the problem with those days?

R/round.r
```diff
@@ -76,25 +77,30 @@ floor_date <- function(x, unit = c("second","minute","hour","day", "week", "mont
...
 	if(!length(x)) return(x)
-  
+
   if (unit == "second") {
     second(x) <- ceiling(second(x))
     return(x)
   }
-  
+
+  if(unit != "quarter") {
```
vspinu: You still have this special treatment of "quarter" here.

R/round.r
jonboiser: Oh wow, I just tried this out:
```
x <- ymd("2014-11-15")
month(x) <- month(x) + 3
x #=>  "2015-02-01 UTC"
```
That's pretty cool. I forget the OO terminology, but is it like `month` is an "active attribute" or something like that, such that if it is changed, the entire object is updated? I had assumed that it would return something like `month(x) = 14` after the update.

vspinu: > So it truncates the mday to one as well?

No. Something is wrong with  your lubridate. 

```
> x <- ymd("2014-11-15")
month(x) <- month(x) + 3
x
 > [1] "2015-02-15 UTC"
```

jonboiser: My mistake. I was looking at the output after changing x to 11/1.

R/round.r
```diff
@@ -159,10 +161,10 @@ parse_unit_spec <- function(unitspec) {
     mult <- as.numeric(parts[[1]])
     unit <- parts[[2]]
   }
-  
+
```
jonboiser: I was looking at one your PRs and Hadley mentioned cleaning up the diffs to not show stuff like this. How do you remove diffs like this which appear from adding/removing newline characters?

vspinu: Seems not to be easy.  http://stackoverflow.com/questions/3515597/git-add-only-non-whitespace-changes

Just don't introduce those in the first place. How did you introduce so many whitespace changes?

I am usually unstaging manually those with `magit` in emacs. You have to figure out how to do that in your front end.

jonboiser: They first showed up when I added a new line to the switch statement in ceiling_date and shifted the contents below it. I've never really thought of these whitespace modifications before, but in RStudio, you can discard the these kinds of chunks so they don't appear in the commit.

Is this PR almost finished? Do the tests and docs need to be modified to reflect this addition?

vspinu: > Is this PR almost finished? Do the tests and docs need to be modified to reflect this 

Unit test fail. 

Github doesn't show your indentation right. It's completely messed up. Are you using TABs to indent your code? Please use simple spaces. R studio must have an option for that. 

Also please squash all the commits into one.  You can add one test for some corner case.

Thanks.

vspinu: > Unit test fail. 

Ok. It's the doc warning. We can leave with this. If it's not difficult, could you please re-roxygenize with the dev version of roxygen2 or adjust the Rd manually.

jonboiser:
Took awhile, but got the PR all cleaned up with a minimal diff.

vspinu:
:+1:  Thanks.

PR #347, opened 2015-09-29
[Fix #336] coercion from interval to period

### LINKED ISSUE #336: Lubridate jumps over a month
Lubridate jumps over the 3ed month here:

as.period(interval(ymd("1985-12-31"),ymd("1986-01-01")))%/%months(1)
as.period(interval(ymd("1985-12-31"),ymd("1986-02-01")))%/%months(1)
as.period(interval(ymd("1985-12-31"),ymd("1986-03-01")))%/%months(1)
as.period(interval(ymd("1985-12-31"),ymd("1986-04-01")))%/%months(1)
as.period(interval(ymd("1985-12-31"),ymd("1986-05-01")))%/%months(1)

 This seems to be a bug, or at least an unexpected calculation.

AlunHewinson:
If the sign of the `@month` and `@day` slots of the new period are opposite to each other, set `@day` to zero.
https://github.com/hadley/lubridate/issues/336

It is my understanding that negative periods are legal but a period created from an interval should not have a negative month slot and positive days or vice versa.

First attempted in pull request https://github.com/hadley/lubridate/pull/346 but there was an error in the test and it didn't handle NAs.  Sorry if it fails the `testthat` tests; I need to learn how to understand and run those myself locally.  Hope you take this attempted bugfix in good faith if it doesn't work; I won't try again if it fails.

### DIFF
--- a/R/coercion.r
+++ b/R/coercion.r
@@ -441,9 +441,10 @@ setMethod("as.period", signature(x = "Interval"), function(x, unit = NULL, ...) 
   to.per$month[nmons] <- 12 + to.per$month[nmons]
   to.per$year[nmons] <- to.per$year[nmons] - 1
   
-  np <- new("Period", to.per$second, year = to.per$year, month = to.per$month, 
+  np <- new("Period", to.per$second, year = to.per$year, month = to.per$month,
             day = to.per$day, hour = to.per$hour, minute = to.per$minute)
-  if (abs(sign(np@month) - sign(np@day))) np@day <- 0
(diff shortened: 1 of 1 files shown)


vspinu:
I don't understand why you are setting day to 0. If the day componnet is negative, it's negative for a reason or not? Something is going wrong with the logic in that function. You cannot just overwrite it with 0 and label as fixed.  

I will try to have a look at it today.

(You don't need close an issue and open a new one each time you change a thing. Simply `git push --force` to the same branch.)

vspinu:
Fixed in 0bbe8c1

PR #351, opened 2015-10-08
origin should be a POSIXct class and inherit from POSIXt

fieldryand:
I think there is just a typo here, because right now you get an error from `origin + milliseconds(1)`


vspinu:
Thanks!

Issue #374, opened 2016-01-22
Messy NEWS


Ironholds:
https://cran.rstudio.com/web/packages/lubridate/news.html is extremely messy for me, because of...something.

Any chance it could be switched out for a plain ol' NEWS file? Not only would this not result in formatting problems, it'd be grabbable from the RStudio updates pane.

vspinu:
News.md is a clean [text file](https://github.com/hadley/lubridate/blob/master/NEWS.md). I guess it's a question for R folks what are they using to convert .md to html. 

@hadley, any ideas of how to fix this? Otherwise I will report an R bug. Using markdown for news file is a standard practice.

Ironholds:
Well, I see problems rendering it with the RStudio preview, so it's not just whatever CRAN is using. It looks like the problem is the ------ above titles; it's reading everything as a title and ignoring the markup below it.

PR #375, opened 2016-01-22
Fix rmd/CRAN issues

Ironholds:
This solves the Rmd/knitr rendering problems at my end; hopefully it works for the CRAN version, too (#374)


vspinu:
How about reporting this to knitr?

vspinu:
Ok. Let's fix that for now. Thanks.

Issue #373, opened 2016-01-21
missing `date` function


joethorley:
lubridate provides `day`, `hour` and `year` functions to get and set the day, hour and year components of a date time. A logical addition would be a `date` function to get and set the date component. It would circumvent the use of as.Date(,tz) which can cause problems if tz is set wrong. It would conflict with the base version of `date` which is poorly named and effectively replaced by lubridate::today. It could be called `d8` to avoid any conflicts. 

Please let me know if I am missing something.

I am willing to implement.

vspinu:
Closely related to #355. 

I would call it  `date` for the sake of consistency. Users of lubridate are very unlikely to use `base::date` anyways.

@hadley, @huftis, what do you think?

hadley:
Hmmm, I'm not a big fan of overriding base R functions (although I agree it wouldn't make much difference here). Why not call it `make_date()` to be consistent with `make_datetime()`?

vspinu:
> Why not call it make_date() to be consistent with make_datetime()?

`make_datetime` is a constructor. The proposal is to add a "getter" and "setter"  for date. That is, to extend the existent family of our accessors -`year`, `day`, `month` etc.

joethorley:
`get_date` and `set_date` could work?

hadley:
Ooooh, I totally misread that

joethorley:
Unless someone tells me otherwise I'm going to do a pull request early next week for `get_date` and `set_date`

vspinu:
`get_date` and `set_date` is pretty bad IMO. It conflicts with R's `foo<-` convention and makes our accessors heterogeneous, which I don't like at all. 

Overwriting `date` is a small price  to pay for clarity and consistency of `date(x)` and `date(x) <- d`.  Let's just go with overwrite for now and change it in a later stage if we happen to have a better idea.

joethorley:
Agreed - I'll implement early next week

Sent from my iPhone

On Jan 22, 2016, at 22:02, Vitalie Spinu <notifications@github.com> wrote:

get_date and set_date is pretty bad IMO. It conflicts with R's foo<-
convention and makes our accessors heterogeneous, which I don't like at
all.

Overwriting date is a small price to pay for clarity and consistency of
date(x) and date(x) <- d. Let's just go with overwrite for now and change
it in a later stage if we happen to have a better idea.

—
Reply to this email directly or view it on GitHub
<https://github.com/hadley/lubridate/issues/373#issuecomment-174152431>.

Issue #355, opened 2015-11-03
Add make_date() feature for converting date-times into dates


huftis:
I propose adding a convenience function `make_date()` for converting date-times to dates. One might by tempted to use `as.Date()`, and this works in many cases, but with the default options it gives surprising results in a few cases. Example:

  x1 = as.POSIXct("2010-08-03 00:59:59.23")
  x2 = as.POSIXct("2010-08-03 00:59:59.23", tz="Europe/London")
  x3 = as.POSIXct("2010-11-03 00:59:59.23")
  x4 = as.POSIXct("2010-11-03 00:59:59.23", tz="Europe/London")

The naïve user might expect `as.Date()` to return the 3rd of the month (August/November) for all of `x1`–`x4`. However, on my system (a `Europe/Oslo` timezone), I get:

    > as.Date(x1)
    [1] "2010-08-02"
    > as.Date(x2)
    [1] "2010-08-02"
    > as.Date(x3)
    [1] "2010-11-02"
    > as.Date(x4)
    [1] "2010-11-03"

It’s even more surprising if you have already used `floor_date(x, "day")` on the objects to remove the ‘time’ part.

The way to get the correct date (`Date`) corresponding to a date-time is either `as.Date(format(x, "%Y-%m-%d"))` or (preferably) `as.Date(x, tz=attr(x, "tzone"))`:

    > as.Date(x1, tz=attr(x1, "tzone"))
    [1] "2010-08-03"
    > as.Date(x2, tz=attr(x2, "tzone"))
    [1] "2010-08-03"
    > as.Date(x3, tz=attr(x3, "tzone"))
    [1] "2010-11-03"
    > as.Date(x4, tz=attr(x4, "tzone"))
    [1] "2010-11-03"

So I propose adding such a convenience function `make_date()` to `lubridate`, that would run

    as.Date(x, tz=attr(x, "tzone"))

for `POSIXct` and `POSIXlt` objects (and just return `x` for `Date()` objects).

I just based the name on the `make_difftime()` function; other names may better.

hadley:
This seems more like a bug in base R, in my opinion.

vspinu:
> bug in base R,

I also think this should be addressed in base R. No need to duplicate existent functions. Would you bother firing a bug report? 

I think `make_date` and `make_datetime` could serve as better names  for existent `ISOdate` and `ISOdatetime` functions.

huftis:
[...]

PR #376, opened 2016-01-26
Added date accessor `date` (Issue #373)

joethorley:


joethorley:
This is my first attempt at a pul request please let me now if I have made any faux pas

vspinu:
> I made the documentation for date these files also got updated.

re-documenting should not touch code files. I think you did some merges locally and some of the changes that I did before now show as your changes. Try to re-base on master and if it doesn't help, start with a clean tree and copy paste your `accessors-date.R` file and re-document again.

vspinu:
We will also need some tests for this functions. Could you please add a few? Also please look at #355. I think using `make_datetime` and then `as.Date` might get us into trouble. 

BTW, the commit message should probably be "[Fix #373] Add 'date' accessors` in order to automatically close the issue.

Thanks.

PR #379, opened 2016-01-28
clean up whitespace

joethorley:
More clean up ready for date


vspinu:
Are you planing to do whitespace cleanups for all files like this? There are ways to clean all files of trailing whitespaces but I don't see much point in that. It's an editor thing. Some editors just leave those spaces after auto-fill, indent etc. Please don't bother with them.

joethorley:
I've switched off in my editor

PR #380, opened 2016-01-30
[Fix #373] Add 'date' accessors

### LINKED ISSUE #373: missing `date` function

joethorley:
> We will also need some tests for this functions. Could you please add a few? Also please look at #355. I think using make_datetime and then as.Date might get us into trouble.

I have added some tests. Also I avoid the problem of using `make_datetime` and then `as.Date` raised in by pasting Year-Month-Day to a character and then using `as.Date`. 

Note the `date` function provides the `make_date` function proposed in issue #355 (which @hadley suggested be called `as_date`)

For more discussion about the `date` and `date<-` functions see #373.


vspinu:
Two notes on working with git:

 - You don't need to open a new issue each time you re-write your commits. Just do `git push --force`. Github will update the pull request automatically.
 - Each commits must deal with one issue. That makes it easy to navigate in the history.  Could you please squash your two commits into one. If you are using R-studio or some other git-aware editor, there must be a way to do an interactive rebase and squash commits interactively. Otherwise from command line follow these [instructions](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit).

joethorley:
Well I squashed but in the process I managed to introduce conflicts! 
Sorry.
I'm going to cancel this pull request and retry fresh based on what I have learned!

PR #383, opened 2016-02-19
Specify GPL version from src/tparse.c

dpocock:
src/tparse.c specifies GPL-2 but the DESCRIPTION file just says GPL.  This patch fixes it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: lubridate
 Maintainer: Vitalie Spinu <spinuvit@gmail.com>
-License: GPL
+License: GPL-2
 Title: Make Dealing with Dates a Little Easier
 LazyData: true
 Type: Package


vspinu:
Thanks.

khotilov:
This change messed up the lubridate licensing. Previously, the package license was simply GPL, without mention of the specific version. IANAL, but my understanding was that it allowed us to [assume any GPL version](https://cran.r-project.org/doc/manuals/R-exts.html#Licensing). Thus, the assumption of GPLv3 was kind of making it compatible with the Apache-2 license of the included cctz code. A GPLv2 work cannot include Apache-2-licensed code (e.g., see [here](https://www.dwheeler.com/essays/floss-license-slide.html) or [here](https://fedoraproject.org/wiki/Licensing:Main#Good_Licenses)). While there seem to exist some exceptions to that, I doubt those would be relevant for an R package. Besides, the src/tparse.c mentions that it is under GPL "either version 2 of the License, or (at your option) any later version", which I would interpret as GPLv2+. So I don't follow the reason given here for the change to GPLv2.

vspinu:
Right. I was not paying enough attention to licensing :/ Will fix all these licencing issues asap.

PR #389, opened 2016-03-11
Date coercion

imanuelcostigan:
Address #355:

- New `as_date` generic 
- Methods for POSIXt, numeric and character vectors (per `as.Date` S3 methods) with sensible defaults
- Generic and methods are documented.

NB:

* Check passes except for one note regarding the following warning message:

```
@details [periods.r#1]: mismatched braces or quotes 
```


R/coercion.r
```diff
@@ -608,3 +608,43 @@ setMethod("as.character", signature(x = "Duration"), function(x, ...){
 setMethod("as.character", signature(x = "Interval"), function(x, ...){
   format(x)
 })
+
+
+#' Change an object to a Date
```
vspinu: It should probably be "convert".

R/coercion.r
```diff
@@ -608,3 +608,43 @@ setMethod("as.character", signature(x = "Duration"), function(x, ...){
...
+#' @param epoch a Date object, or something which can be coerced by
+#'   \code{as.Date(epoch, ...)} to such an object (default: the Unix epoch of 1
+#'   Jan 1970 as a \code{Date} object). Note that in this instance, \code{x} is
+#'   assumed to reflect the number of days since \code{eoch} at \code{"UTC"}.
+#' @param fmt a character string (default: \code{\%Y-\%m-\%d}). See
+#'   \code{\link{strptime}} for alternative format specifications.
+#' @param ... further arguments to be passed to specific methods (see above).
+#' @return a vector of \code{\link{Date}} objects corresponding to \code{x}.
+#' @export
+setGeneric(name = "as_date", function(x, ...) standardGeneric("as_date"))
```
vspinu: Please add default function to fall back on as.Date.

R/coercion.r
```diff
@@ -608,3 +608,43 @@ setMethod("as.character", signature(x = "Duration"), function(x, ...){
...
+#' @rdname as_date
+#' @export
+setMethod(f = "as_date", signature = "POSIXt", function (x, tzone = tz(x)) {
+  as.Date(x, tz = tzone)
+})
+
+#' @rdname as_date
+#' @export
+setMethod(f = "as_date", signature = "numeric",
+  function (x, epoch = as.Date("1970-01-01")) {
```
vspinu: Why no use "origin" as an argument?

R/coercion.r
```diff
@@ -608,3 +608,43 @@ setMethod("as.character", signature(x = "Duration"), function(x, ...){
...
+#' @param tzone a time zone name (default: time zone of the POSIXt object
+#'   \code{x}). See \code{\link{olson_time_zones}}.
+#' @param epoch a Date object, or something which can be coerced by
+#'   \code{as.Date(epoch, ...)} to such an object (default: the Unix epoch of 1
+#'   Jan 1970 as a \code{Date} object). Note that in this instance, \code{x} is
+#'   assumed to reflect the number of days since \code{eoch} at \code{"UTC"}.
+#' @param fmt a character string (default: \code{\%Y-\%m-\%d}). See
+#'   \code{\link{strptime}} for alternative format specifications.
+#' @param ... further arguments to be passed to specific methods (see above).
+#' @return a vector of \code{\link{Date}} objects corresponding to \code{x}.
```
vspinu: Could you please add a couple of words on the motivation. Why did we add this functions and how it compares to base as.Date?

vspinu:
As to the commiting style, could you please keep related changes in one commit whenever possible? No need to split out documentation, roxigenization and news addition steps. All of these are a single change - adding one feature.

Also could you please add a test case as per #355?

vspinu:
Hi Imanuel. What is the status of this? Do you plan to address my other comments and add tests? I would like to release by the end of this week. I can merge what you have so far and amend those myself if you don't have time.

imanuelcostigan:
@vspinu I'm unlikely to get time before end of the week. In terms of commit style - sure. Suggest that you consider adding a [CONTRIBUTING file to repo with this and other expectations](https://github.com/blog/1184-contributing-guidelines).

vspinu:
Ok. I will then merge manually your changes these days then leave for another shelve for 5-6 days before the final release. Will add the CONTRIBUTING file. Thanks for the idea.

imanuelcostigan:
@vspinu found myself with some time tonight. Have added test cases and default method as requested.

vspinu:
There are merge conflicts. I am merging it right now. By "default" I meant to use `useAsDefault` argument to `setGeneric`, but I guess this doesn't make any difference.

BTW, why do you need a method for "character"? as.Date does good job there. No?

imanuelcostigan:
`as.Date` fine in that respect, but why not give users the same function call for everything? Of course, there is a performance hit by wrapping as.Date in an S4 method call, but I think users will appreciate unified call semantics. If they want performance, they can always call as.Date directly.

vspinu:
Performance is not an issue. We are defaulting to as.Date anyways. So character input will drop into `as.Date.character` which seem to be doing a bit more than format = "%Y-%m-%d".

imanuelcostigan:
Btw see https://github.com/blog/2141-squash-your-commits for setting your repo up to squash commits rather than requiring contributors to do so

PR #409, opened 2016-04-30
Fix period format

alberthkcheng:
```
> period(second = 0.02)  
[1] "2S"
```

The regex omitted the decimal place


vspinu:
Thanks!

Issue #268, opened 2014-10-16
Add options "quarter","bimonthly", and "half-year" to floor_date/ceiling_date [Feature request]


mmeierer:
Add options "quarter","bimonthly", and "halfyear" to floor_date/ceiling_date. 

Thus any given date is rounded up/down to the first day of the quarter, bimonthly period (Jan-Feb, Mar-Apr, May-Jun, Jul-Aug, Sep-Oct, Nov-Dec) or half-year.

jonboiser:
I would be interested in this feature too. Has there been any discussion on how it might be implemented? Are the boundaries of quarters simply 1/1, 4/1, 7/1, 10/1? Same with bimonthly (e.g. 1/1, 3/1, 5/1 ..., 11/1)?

ptoche:
you know about ``round_date(x, "quarter")`` right?

jonboiser:
I submitted the PR #303  that implemented the quarter-round/floor/ceil functionality. I believe this is still open since bimonthly, and halfyear are still unimplemented. This issue hasn't been touched in awhile, but I can look into adding those two missing pieces.

vspinu:
> I can look into adding those two missing pieces.

Yes. Please. It would be useful indeed.

PR #414, opened 2016-05-09
Add "bimonth" and "halfyear" options to round_date and friends

jonboiser:
For #268. Did a little refactoring to re-use the formula used in rounding to "quarter".

Also, deleted that `parse_unit_spec` function since it really was dead code, and changed an errant tab to spaces.

Still need to write specs.

### DIFF
--- a/R/round.r
+++ b/R/round.r
@@ -1,7 +1,7 @@
 #' Round, floor and ceiling methods for date-time objects.
 #'
 #' Users can specify whether to round to the nearest second, minute, hour, day,
-#' week, month, quarter, or year.
+#' week, month, bimonth, quarter, halfyear, or year.
 #'
 #' \code{round_date} takes a date-time object and rounds it to the nearest
 #' integer value of the specified time unit. For rounding date-ties which is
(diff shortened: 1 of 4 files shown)


vspinu:
Thanks. Could you please add tests for this new feature? Also the current checks fail.

jonboiser:
Just added tests for the new options, squashed the commits, and made sure CI passed. Let me know this needs anything else before it's ready.

R/round.r
```diff
@@ -1,7 +1,7 @@
 #' Round, floor and ceiling methods for date-time objects.
 #'
 #' Users can specify whether to round to the nearest second, minute, hour, day,
-#' week, month, quarter, or year.
+#' week, month, "bimonth", quarter, half-year, or year.
```
vspinu: Why is "bimonth" quoted here? half-year should probably be "halfyear".

vspinu:
Very nice!

Does this work with Date objects as expected? I know we don't many tests for that but could you add a couple of checks for Date objects as well? You can produce dates directly with `ymd` or `make_date` now.

Please also add a line to news.md with a link to #268. 

This completely fixes #268, right? If so, please add `[Fix #268]` to the commit header to automatically close it. 

Thanks a lot!

R/round.r
jonboiser: I quoted "bimonth", because I don't think there is a proper term for "two-month period" and "bimonth" is just a made-up word. Same with "halfyear". It might be better though if the terms were consistent with the options.

jonboiser:
All done! I added the extra tests to the "floor/ceiling_date works for a variety of formats" to test that they work on Date classes.

jonboiser:
Had to do a couple more pushes to get the commit message right and rebuild the docs!

vspinu:
Perfect. Thanks!

PR #427, opened 2016-07-05
Add shift parameter to quarters function when fiscal does not match calendar year

PedramNavid:
A calendar year starts in January, but many places have fiscal years that start either before or after the calendar year. For example, a fiscal year might start in November, which means Q1 is Nov - Jan and Q4 is Aug - Oct.

The shift parameter causes two major changes, one is that the shift changes the quarter for a particular month, as described above. 2nd change is that it also increases the year by 1 if the quarter would 'normally' be Q4 but is now Q1 due to a shift. E.g. December 2015 is normally 2015.Q4 but would be reported as 2016.Q1 if the Fiscal Year started in December. 

Created new test cases as well, as I could not find any for the quarter accessor.

### DIFF
--- a/R/accessors-quarter.r
+++ b/R/accessors-quarter.r
@@ -1,17 +1,22 @@
 #' @include accessors-month.r
 NULL
 
-#' Get the fiscal quarter of a date-time.
+#' Get the calendar or fiscal quarter of a date-time.
 #'
-#' Fiscal quarters are a way of dividing the year into fourths. The first quarter (Q1)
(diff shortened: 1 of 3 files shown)


vspinu:
Sorry, I missed this pull somehow. 

Could we find a better name for "shift" ?  It's not imidiately clear which direction it shifts. For instance are `shifts` of -2 and 10 equivalent?

Would we be better off with `year_start` parameter instead? You pass 10 or "November" and get the same result?

PR #446, opened 2016-07-28
Implement as_datetime

### LINKED ISSUE #437: as_datetime()
For consistency with `as_date()`

hadley:
Let me know if you want me to also update the styles of the other generics/methods. (Using named arguments is particularly annoying for me because it breaks function lookup in RStudio :/)

Fixes #437


vspinu:
Thanks! Of course, any cleanups and code simplifications are always welcome. 

I am afraid I don't quite follow what exactly you want to fix with respect to "named arguments".  Do you say that defGeneric should not contain named arguments?

hadley:
Style wise I think that:

```R
setMethod("as_date", "numeric",
  function (x, origin = lubridate::origin) {
    as.Date(x, origin = origin)
   }
)
```

is better than

```R
setMethod(f = "as_date", signature = "numeric",
            function (x, origin = lubridate::origin) {
              as.Date(x, origin = origin)
            })
```

PR #449, opened 2016-07-28
Improve aliasing

hadley:
A whole bunch of small fixes to better generate usage statements


NAMESPACE
```diff
@@ -192,14 +192,6 @@ export(period)
 export(period_to_seconds)
 export(picoseconds)
 export(pm)
-export(pretty.day)
```
krlmlr: Did you intend to remove these exports?

hadley: Yes - they're still exported (correctly) as S3 methods

vspinu:
Looks good, but generated a bunch of doc warnings. Unless you have time to look at it  I will merge it as is and will fix those myself tomorrow.

vspinu:
I cannot find the reason for this one check failure introduce by S3 export refactoring:  

```R
R CMD check results
1 error  | 2 warnings | 0 notes
checking examples ... ERROR
Running examples in ‘lubridate-Ex.R’ failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: pretty_dates
> ### Title: Computes attractive axis breaks for date-time data
> ### Aliases: pretty_dates
> ### Keywords: chron dplot utilities
> 
> ### ** Examples
> 
> x <- seq.Date(as.Date("2009-08-02"), by = "year", length.out = 2)
> # "2009-08-02" "2010-08-02"
> pretty_dates(x, 12)
Error in get(as.character(FUN), mode = "function", envir = envir) : 
  object 'pretty.month' of mode 'function' was not found
Calls: pretty_dates -> match.fun -> get
Execution halted
```

Any ideas?

hadley:
I see what the problem is - I'll fix it later in the week (currently busy at a conference)

hadley:
Should be good now.  The subcomponents of pretty are no longer exported, which is technically a breaking change. But it seems unlikely that they'd be used by anyone.

PR #455, opened 2016-08-01
Fix decimal_date and date_decimal handling of daylight savings time b…

### LINKED ISSUE #450: decimal_date doesn't handle daylight savings time boundaries.
The decimal_date function doesn't seem to handle daylight savings time boundaries correctly. 
```
> date4 <- ymd_hms("2016-03-13 01:00:00", tz = 'America/New_York')
> date4        <- date4 + dhours(-1:1)
> decimal4     <- decimal_date(date4)
> diff(decimal4)
[1] 0.0001138434 0.0002276867
```
The times above are a fixed hour apart however the decimal_date function ignores the tz information. Not sure this is intentional but it doesn't seem necessary unless I'm missing something.

I've fixed the issue in my fork and and wrote a couple tests. All tests in test-decimal-date passed. I'll issue a pull request shortly.

jasonelaw:
…oundaries. In coercing to POSIXlt and using force_tz, the original code did not respect daylight savings time shifts. This fixes that issue and and adds tests for the functionality. Closes #450

### DIFF
--- a/R/decimal-dates.r
+++ b/R/decimal-dates.r
@@ -55,4 +55,5 @@ date_decimal <- function(decimal, tz = "UTC") {
   seconds <- as.numeric(difftime(end, start, units = "secs"))
   frac <- decimal - Y
   end <- start + seconds*frac
+  return(end)
 }
--- a/R/decimal-dates.r
+++ b/R/decimal-dates.r
@@ -16,10 +16,9 @@ decimal_date.default <- function(date){
   if(any(!inherits(date, c("POSIXt", "POSIXct", "POSIXlt", "Date"))))
     stop("date(s) not in POSIXt or Date format")
 
-  date <- force_tz(as.POSIXlt(date), tzone = "UTC")
   Y <- year(date)
-  start <- make_datetime(Y, 1L, 1L)
-  end <- make_datetime(Y + 1L, 1L, 1L)
+  start <- make_datetime(Y, 1L, 1L, tz = tz(date))
+  end <- make_datetime(Y + 1L, 1L, 1L, tz = tz(date))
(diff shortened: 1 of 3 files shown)


R/decimal-dates.r
```diff
@@ -48,16 +47,12 @@ decimal_date.its <- function(date)
...
   Y <- trunc(decimal)
   ## parsing is much faster than updating
-  start <- make_datetime(Y, 1L, 1L)
-  end <- make_datetime(Y + 1L, 1L, 1L)
+  start <- make_datetime(Y, 1L, 1L, tz = tz)
+  end <- make_datetime(Y + 1L, 1L, 1L, tz = tz)
   seconds <- as.numeric(difftime(end, start, units = "secs"))
   frac <- decimal - Y
   end <- start + seconds*frac
-  if (!is.null(tz))
```
vspinu: Why did you remove this? `force_tz` is very heavy and is not needed for `UTC`.

vspinu:
Thanks for this. The doc checks are failing because you have renamed the argument and didn't re-run devtools::document. Could you please re-base and fix that?

R/decimal-dates.r
vspinu: Nevermind. That removal is the main point of the RP. Could you please explicitly return the value? Now the return value is invisible.  Thanks.

jasonelaw:
@vspinu, how does this look? I'm no git expert so I didn't attempt to clean up the commits. They're a little messy.

vspinu:
Yeh, a bit unclean but it's fine. Also a news entry should have been added. Nevermind, I will fix that.

Issue #459, opened 2016-08-17
Finance use case - tenor conversions to time spans


waynenilsen:
I thought I would add this as an issue, hoping that some functionality like this is already in lubridate. I have what we call tenors in finance, examples are `1w` `5d` `10y` etc. I think the `months` and `days` function serve this purpose however considering the ease-of-use nature of this library a convenience function would be great for converting these number-plus-span combinations into lubridate things. Let's call this function `tenor` for the sake of argument. The nature of this function would be such that

```
tenor('1d') = days(1)
tenor('1w') = weeks(1)
tenor('1m') = months(1)
tenor('1q') = quarters(1)
tenor('1y') = years(1)
```

Is there already something like this?

vspinu:
Do I understand correctly that you want specialized parsers to get from `3d` into `days(3)`? 

`tenor` is obscure. I would better generalize the `period` constructor to be able to do `period("2d")` for this.

There is an internal unit parser since in lubridate:

```R
> lubridate:::parse_period_unit("1y")
$n
[1] 1

$unit
[1] "y"
```
That's the closest it gets at the moment.

waynenilsen:
You understand the concept

The name I leave to you, period seems like a reasonable name for it. The `parse_period_unit` function would be helpful but it is not exposed.

Implementation could be along the lines below? I can take a stab if you want although I have never really gotten into R package dev before not sure about testing.

```

function_mapping = list(
	"Y" => years,
	"M" => months,
	...
)

period = function(spec) {
	parsed = parse_period_unit(spec)
	function_mapping[[parsed$unit]](parsed$n)
}
```

vspinu:
> Were you thinking of adding another case here?

Yes. 

But let's think of a proper and consistent API first. I would very much prefer to have a more generic string parser. Something like `period("1d 2min") should also work. Pretty much like with current `period(days=1,months=2)` but within string.

Also need to check what ISO standard has to say about it. Related #362

waynenilsen:
I certainly agree that parsing the ISO standards is an excellent goal to reach for.

My question is, how about both? That is, have the ability to parse simple `1y` as well as the full ISO specification. Perhaps the `period` function should parse according to ISO and this parsing could be in a separate function. 

Of course, if you would rather this functionality be external to lubridate, just let me know and I will close this issue and related PR.

vspinu:
>  how about both?

Yes. This is the intention. I don't think it should be external to lubridate.  

[...]

PR #460, opened 2016-08-17
string argument support in period function

waynenilsen:
Work in progress pr for #459 

I could not figure out how to actually run the test but I did write the test.


R/periods.r
```diff
@@ -465,6 +476,12 @@ months.numeric <- function(x, abbreviate) {
   period(month = x)
 }
 
+.period_mapping = list(
```
waynenilsen: I think this makes sense but I'm not sure about R scoping rules and how long this object will live for..

vspinu:
parse_period_unit works on a vector of length one by design. For what you propose function must be vectorized. 

One way is to go with relatively complex multi regexp matcher (see regexpr function in R) with perl style captures. An alternative is to implement the parser in C directly. I am favoring the latter because is likely to be more reliable and easier to extend.

PR #465, opened 2016-09-14
Fix format.Duration for durations greater than a minute with fractional seconds

tsbattman:
Previously format(duration(75.25)), for example, would fail because .read_duration used %d as the format specifier in sprintf. In order to match the existing format.Duration tests, I simply round the seconds.

Added tests to test-durations.R

### DIFF
--- a/R/durations.r
+++ b/R/durations.r
@@ -70,7 +70,7 @@ SECONDS_IN_ONE <- c(
     paste0(x, "s")
   else {
     x2 <- round(x / SECONDS_IN_ONE[[unit]], 2)
-    sprintf("%ss (~%s %ss)", x, x2, unit)
+    sprintf("%ds (~%s %ss)", round(x), x2, unit)
   }
 }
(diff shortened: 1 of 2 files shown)


vspinu:
I think we need to show the fractional seconds then. `round(x)` doesn't look like a correct solution. I think we can safely use %s format there.

vspinu:
Would you like to adjust your PR for %s?

tsbattman:
Sure deal. Should have it within a day.

vspinu:
@tsbattman is this ready to be pushed?

tsbattman:
I put in the change to use %s and modified the test suite accordingly. Should be good to go.

vspinu:
Thanks!

wlandau-lilly:
I am glad this was fixed. When do you think we might see an update on CRAN?

vspinu:
Time permitting, next week.

PR #480, opened 2016-09-27
fix ceiling_date return type when unit = "week"

### LINKED ISSUE #479: wrong return type in `ceiling_date` when unit = "week"
Hi, @vspinu, I tried the CRAN version of `lubridate` and found a bug related to `ceiling_date` when the `unit = "week"`. It should return a date object instead of a datetime. 

Here's the reproducible example:

```r
library(lubridate)

# ceiling date returns a time
ceiling_date(ymd("20160927"), "week")
## [1] "2016-10-02 08:00:00 CST"

# while the floor_date returns a date
floor_date(ymd("20160927"), "week")
## [1] "2016-09-25"
```

shrektan:
Closes #479

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,7 @@ Version 1.6.0
 
 ### BUG FIXES
 
+* [#479](https://github.com/hadley/lubridate/issues/479) Fix the inconsistent behavior in `ceiling_date` when `unit = "week"`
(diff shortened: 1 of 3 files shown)


vspinu:
Could you please add a test or two and add a line to the news file? Thanks.

shrektan:
I've added tests and lines in news. You're welcome. 😃

PR #485, opened 2016-10-21
add fiscal_start to quarter when fiscal year doesn't begin in january

PedramNavid:
This is a follow-up to Pull Request #427 , but I couldn't figure out how to use git properly to bring my old fork up so I deleted and started again. 

I removed shift as requested, and created a fiscal_start parameter instead. Specifying a month other than 1 for the start of the fiscal year will shift the quarter (and the year if needed). For example, setting 11 for start will mean that Nov, Dec, Jan are all Q1. For Nov and Dec, the year will increment by 1 if with_year = TRUE. That is, 2013-11 will show as 2014.1 (1st quarter of 2014, as is convention when using fiscal quarters). 

I have also added tests for quarters which did not exist previously.

### DIFF
--- a/R/accessors-quarter.r
+++ b/R/accessors-quarter.r
@@ -10,19 +10,28 @@ NULL
 #'   anything else that can be converted with as.POSIXlt
 #' @param with_year logical indicating whether or not to include the quarter's
 #'   year.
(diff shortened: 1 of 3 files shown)


vspinu:
Looks good. Thanks!

PR #508, opened 2017-01-16
[Fix #401 ext] Return localized labels in month

### LINKED ISSUE #401: Return localized weekdays for wday()
wday() always returns English names of weekdays. I made the following changes using date_names_lang in readr, to overcome this. Is there solution exists without using readr?

Also, the same applies to month.

```
wday.numeric <- function (x, label = FALSE, abbr = TRUE, language="en") 
{
  if (!label) 
    return(x)
  if (abbr) {
    labels <- date_names_lang(language)$day_ab
  }
  else {
    labels <- date_names_lang(language)$day
  }
  ordered(x, levels = 1:7, labels = labels)
}

wday(ymd(160408),label=TRUE,language="ko")
# [1] 금
# Levels: 일 < 월 < 화 < 수 < 목 < 금 < 토
```

alberthkcheng:
Return localized month in the same way as weekday.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@ Version 1.6.0.9000
 ==================
 
 ### NEW FEATURES
+* [#508](https://github.com/hadley/lubridate/pull/508) New parameter `locale` in `month`. Labels of the returned factors (when `label=TRUE`) now respect current locale.
 * [#485](https://github.com/hadley/lubridate/pull/485) `quarter` gained a new argument `fiscal_start` to address the issue of different fiscal conventions.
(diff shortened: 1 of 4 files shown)


alberthkcheng:
pls check again the comment in #401 . thanks :)

vspinu:
Thanks, but this is not backward compatible. But I guess it's the right way to go anyways. 

Could you please add a note to NEWS.md and check if there are tests for english locale?

alberthkcheng:
NEW.md and tests are added for month and wday

PR #525, opened 2017-03-08
[Fix #517] `day<-`, `minute<-`, etc. should not produce an error when…

### LINKED ISSUE #517: assign value using day(), minute(), etc., will yield error when providing empty input
Example as below:

>\> a <- c(Sys.Date(), Sys.Date()+1)
\> a
[1] "2017-02-09" "2017-02-10"
\> day(a[c(F,F)]) 
integer(0)
\> day(a[c(F,F)]) <- 10
Error in data.frame(..., check.names = FALSE) : 
  arguments imply differing number of rows: 0, 1

Whereas in normal R functions, it should just do nothing and do not yield an error:

>\> b <- 1:2
\> b[c(F,F)]
numeric(0)
\> b[c(F,F)] <- 10
\> b
[1] 1 2

timgoodman:
… length(x) is zero


vspinu:
Thanks!

PR #537, opened 2017-04-26
Convert documentation to Markdown

krlmlr:
Also fixed a few typos.

PR #538, opened 2017-04-26
New time_of_day() and as.hms() for timespan classes

krlmlr:
`time_of_day()` returns an `Interval`, `as.hms()` forwards to `as.difftime()`.

Closes #537 (=includes it).

Fixes tidyverse/hms#28.

@hadley: Could you please take a look?


randomgambit:
Hello there, following some interesting discussion with @krlmlr (https://github.com/tidyverse/hms/issues/32), I post here my feedback on these changes (hoping that helps)

I see @krlmlr's point of keeping hms as lightweight as possible, but consider the very frequent setting where someone works with intraday timestamped data, and needs to filter data between some business hours every day. Using hms(timestamp) > hms('09:30') is a natural way to do so. The horrible way to do so is to extract the time part of a timestamp (in string format) and compare to > '09:30:00'

Right now, `lubridate` does not have a good support for time of day and that is why I think both packages are complementary. Ultimately, they could be combined into one (in the same fashion as in Pandas) but this is a matter of taste

vspinu:
`as.hms` is rather confusing in the lubridate context. We have `hms` parser which returns a period. In retrospect it should have probably been duration, but now it's too late. 

Also `time_of_day` should return a duration not an interval. The best way to achieve that is to invoke `as.POSIXlt` and extract hour, min and sec from there, no floor_date involved. This is pretty lightweight and can be in the hms package itself. 

I also think  `as.hms` is unnecessary. You can make the constructor hms dispatch on the object. No need for `time_of_day` either, `hms(posix_object)` can do that. This would address the use cases by @randomgambit, if I am not mistaken.

randomgambit:
@vspinu thanks for your answer. 

I am not quite sure to undestand what you propose.

Say I have a string: `timestamp = '2015-05-05 20:02:42.453`' that I can parse using
` time_parsed = ymd_hms(timestamp)`

What you are saying is that we can type directly
`hms(time_parsed)`  without having to use the `as.hms` syntax?

krlmlr:
`hms(Sys.time())` doesn't work on my system:

``` r
lubridate::hms(Sys.time())
#> Warning in .parse_hms(..., order = "HMS", quiet = quiet): Some strings
#> failed to parse
#> [1] NA
```

Thanks for your suggestion to use `POSIXlt`, there seems to be a use case for that data type after all ;-)

vspinu:
> hms(Sys.time()) doesn't work on my system:

`hms` is a parser, it accepts strings only. We can generalize it of course, but it will return a period (not a duration or difftime). A period is a pretty heavy object and doesn't play nicely with data frames. This is precisely why I am in favour to keep hms lightweight objects separately.

vspinu:
> What you are saying is that we can type directly hms(time_parsed) without having to use the as.hms syntax?

No. I was proposing to @krlmlr to avoid `as.hms` altogether in `hms` package but to generalize `hms::hms` constructor to accept strings, POSIXt and Date objects. So that you can do `hms(timestamp) > hms("9:30")` without much ado. Of course that would mean loading hms after lubridate because `hms` has a different meaning in lubridate.

PR #549, opened 2017-05-31
Fixed typo in hms man description

cmartin:


vspinu:
Thanks. But it should be fixed in the code's roxygen comment. Rd's are product of roxygenition.

PR #550, opened 2017-05-31
Fix a typo in hms documentation

cmartin:
This is the second pull request for this problem, now modifying the roxygen documentation instead of the resulting Rd file.

Sorry for the trouble...


vspinu:
Thanks.

PR #554, opened 2017-06-16
Fix tests when run under non English locales:

zeehio:
The testsuite was failing on a computer with a non-English default locale. This PR fixes the tests so they work well under any locale.

- days_in_month: As N_DAYS_IN_MONTHS has English month names, month() needs to be called with an English locale (the "C" locale is English and always available)

- wday test: The wday in the test is compared with the English week day, so needs to be in English

- "month accessor extracts correct month": Comparison with English month names, needs C locale as well


vspinu:
Thanks!

PR #577, opened 2017-09-21
styling with styler

lorenzwalthert:
This PR applies the [styler](https://lorenzwalthert.github.io/gsoc2017/) package to format the code according to the [tidyverse](http://style.tidyverse.org/) style guide. Hadley Wickham asked me and @krlmlr to [apply the formatter to some tidyverse packages](https://twitter.com/hadleywickham/status/904461123684007936). I ran 
```r
styler::style_pkg(scope = "spaces", strict = FALSE")
```
And added some tweaks  - as styler is not (yet) perfect. This is the least invasive configuration so I felt it might be a good place to start and it's also easier to review the application of different levels of invasiveness separately. Maybe we can do that once this PR is approved? Let me know what you think and/or if you see any malformatting introduced with this PR.


R/difftimes.r
```diff
@@ -79,13 +79,13 @@ make_difftime <- function(num = NULL, units = "auto", ...) {
   units <- standardise_date_names(units[[1]])
   switch(units,
          second = structure(num, units = "secs", class = "difftime"),
```
krlmlr: Was this indention a manual tweak?

lorenzwalthert: No, since I used `scope = "spaces`, indention was not touched.

lorenzwalthert:
Right... I guess it would not be that hard. However, I thought the idea of the tidyverse style guide was to have all packages formatted as uniformly as possible.

vspinu:
I personally quite disagree with tidy's style on formating of divisions and multiplications.

```R
     inc_year <- q == 1 & uq == 4
-    year(x) + inc_year + q/10
+    year(x) + inc_year + q / 10


-  if (with_year) year(x) + s/10
+  if (with_year) year(x) + s / 10

-  Y + sofar/total
+  Y + sofar / total
```

I don't think it's natural (as in hand-writing) nor it increases readability. So I would prefer to apply this without those changes. Is there an option in styler for this level of detail? If not, then be it.

lorenzwalthert:
There is an open issue on more granular control of math token spacing (https://github.com/krlmlr/styler/issues/86). From a technical point of view, it should not be difficult to implement that.

lorenzwalthert:
I used krlmlr/styler#221 which aims to solve krlmlr/styler#86 to style the package again, so spacing is should be set as desired.
I ran
```r
style_pkg(
  strict = FALSE, 
  scope = "spaces", 
  math_token_spacing = specify_math_token_spacing(zero = c("'/'","'*'"))
)
```

To review the relevant diffs more easily, I used `git diff e791315 25be8a7`.

vspinu:
Thanks! :+1:

Issue #499, opened 2016-12-14
problem with loadMethod for lubridate in Rscript


pchtsp:
I had written in #314 but, since it's already close, I think it's better to open a new one. Although *it seems* to be the same problem. Also similar to #407, apparently.

This (original example) works:

    Rscript -e "lubridate::round_date(Sys.time(),'minute')"

But this one doesn't:

    Rscript -e "lubridate::dhours(22)"

I get the following (sorry for the spanish):

    Error in .setupMethodsTables(fdef, initialize = TRUE) :
      tentativa de obtener un slot "group" de un objeto de una clase básica ("NULL") sin slots
    Calls: <Anonymous> ... getMethodsForDispatch -> .getMethodsTable -> .setupMethodsTables
    Ejecución interrumpida

I'm using lubridate as dependency in a package I'm building. I'm calling the package from a single R script that gets executed via Rscript.
I have tried adding `methods` and or `lubridate` as imports in the NAMESPACE of the package and still get the same error. I have had to add `library(methods)` to the R script that calls the package in order to make it work. But this doesn't seem to be a "nice" solution to the problem.

Any advice?
Thanks!

vspinu:
Which version of R is this? I am getting this on 3.2.4:

```r
 Rscript -e "lubridate::dhours(22)"
Loading required package: grDevices
[1] "79200s (~22 hours)"
Warning message:
In .doLoadActions(where, attach) :
  trying to execute load actions without 'methods' package
```
warning is annoying but there is little we can do about it (I think).

pchtsp:
sorry, forgot to add sessionInfo(), here it goes:

    R version 3.3.1 (2016-06-21)
    Platform: x86_64-w64-mingw32/x64 (64-bit)
    Running under: Windows 10 x64 (build 14393)

    locale:
    [1] LC_COLLATE=Spanish_Spain.1252  LC_CTYPE=Spanish_Spain.125
    [3] LC_MONETARY=Spanish_Spain.1252 LC_NUMERIC=C
    [5] LC_TIME=Spanish_Spain.1252

    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  base


I tried also with `R-3.2.3`, which I also had installed in my pc. I only got the warning as you mentioned.

edzer:
`Rscript -e`, as opposed to `R -e`, does not load `methods` on startup. If you need to use `Rscript -e`, load it manually:
```r
[...]

Issue #407, opened 2016-04-28
`methods` not loaded/working when using lubridate::object notation in Rscript


mmuurr:
I'm having some trouble executing functions that rely on package **methods** when being run via Rscript. I already know that **methods** is not _attached_ via Rscript, but it should be loaded (and usable) by the dependent packages and accessible via their package namespaces, at least I thought so...

Example R script (executed via `Rscript`):

    #!/usr/bin/Rscript
    
    sessionInfo()
    
    tryCatch({
        str(lubridate::days(30))
    }, error = function(e) {
        print(e)
    })

    sessionInfo()

    tryCatch({
        str(lubridate::days(30))
    }, error = function(e) {
        print(e)
    })
    
    library(methods)
    str(lubridate::days(30))
    sessionInfo()

Output, with my comments and questions mixed below ...

First, here's the startup `sessionInfo()`:

    # sessionInfo()

    R version 3.2.2 (2015-08-14)
    Platform: x86_64-redhat-linux-gnu (64-bit)
    Running under: Amazon Linux AMI 2015.09
    
    locale:
     [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
     [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8
     [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8
     [7] LC_PAPER=en_US.UTF-8       LC_NAME=C
     [9] LC_ADDRESS=C               LC_TELEPHONE=C
    [11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C
    
    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  base

As expected with `Rscript`, package **methods** is not attached, nor loaded.

Now, what's surprising to me is that I thought the `lubridate::days(30)` call would load **methods**, since the **lubridate** package does "Depend" on **methods**, as verified in **lubridate**'s `DESCRIPTION` file. Instead, I get the error:

    # tryCatch({
    #     str(lubridate::days(30))
    # }, error = function(e) {
    #     print(e)
    # })

    <simpleError in .setupMethodsTables(fdef, initialize = TRUE): trying to get slot "group" from an object of a basic class ("NULL") with no slots>

But now I'm _really_ confused, because **methods** _does_ seem to have been loaded, as seen with the next `sessionInfo()` call:
  
    # sessionInfo()  
  
[...]

Issue #314, opened 2015-04-22
problem with loadMethod for lubridate in Rscript


mccreigh:
Compose the Rscript with file name `lubridateTest.Rsh`:

```{r}
#! /usr/bin/RScript
lubridate::round_date(Sys.time(),'minute')
```
Run it:
```{bash}
jamesmcc@chimayo:~> ./lubridateTest.Rsh 
Error in x - eseconds(1) : could not find function "loadMethod"
Calls: <Anonymous> ... as.POSIXct -> ceiling_date -> floor_date -> update -> -
recover called non-interactively; frames dumped, use debugger() to view
```
This works however:
```{r}
#! /usr/bin/RScript
library(lubridate)
lubridate::round_date(Sys.time(),'minute')
```
Run it:
```{r}
jamesmcc@chimayo:~> ./lubridateTest.Rsh 
Loading required package: methods
[1] "2015-04-22 13:25:00 MDT"
```

I have no idea why, but it just is this way. I'm curious to know however.
Thanks,
J

vspinu:
This applies to all S4 packages. The problem is that `methods` package is not attached from `Rscript`. This will work:

```
#! /usr/bin/Rscript
library(methods)
lubridate::round_date(Sys.time(),'minute')
```

I think there is a way around it, like explicitly calling `require(methods)` in `evalOnLoad`, but I am not sure we should bother with that. @hadley should know better what is the recommended way to deal with this issue.

hadley:
I think this might be fixed in 3.2

vspinu:
It's still there in today's devel but under a different form:

```
Error in dseconds(1) : could not find function "new"
Calls: <Anonymous> ... as.POSIXct -> ceiling_date -> floor_date -> update -> dseconds
In addition: Warning message:
no function found corresponding to methods exports from ‘lubridate’ for: ‘show’ 
Execution halted
```

hadley:
Are you importing the methods package and new etc?

vspinu:
> Are you importing the methods package and new etc?

Nope. `methods` is in Depends and no explicit imports for `new`, `show` etc are there. I guess that was it.

vspinu:
Ok. I have imported `new` and `show`. Now I am getting

```
Error in .setupMethodsTables(fdef, initialize = TRUE) : 
  trying to get slot "group" from an object of a basic class ("NULL") with no slots
[...]

PR #583, opened 2017-10-06
[#407, #499] Import from methods 

infotroph:
# Motivation

This patch tries to resolve several cases where lubridate does not work reliably when methods is loaded but not attached. In previous discussions (#314, #407, #499, ...) the sense has been that this is R's fault rather than lubridate's. While I agree that having `Depends: methods` in the package description *ought to be* sufficient to ensure that methods is available, experience shows that it is not and I vigorously disagree that "make the user load methods for themselves" is a solution. See also the [R package guide](http://r-pkgs.had.co.nz/namespace.html#imports): "If your package is loaded, packages in Depends are loaded, but not attached, so you need to qualify function names with :: or specifically import them."

The current behavior leaves me, like previous reporters, in a bind because I want to use lubridate in my own packages without extending the cascade of `Depends`. The [general consensus](https://stackoverflow.com/questions/8637993/better-explanation-of-when-to-use-imports-depends) for what to do in this situation seems to be ["Contact the maintainer and ask them to do a more careful job of constructing their namespace."](https://stackoverflow.com/questions/14808337/tmplot-is-in-namespace-but-its-dependencies-are-not-found/14813080#comment20749830_14813080), so... hi, maintainers. Does it help that I come with a patch in hand?

# Contents

This PR follows the same approach as #314 -- it explicitly `@import`s the relevant functions (`initialize`, `coerce<-`) from `methods`. 

I gather that the two unresolved issues in #314 were [how to know what to import](https://github.com/tidyverse/lubridate/issues/314#issuecomment-101763989) and [how to test it effectively](https://github.com/tidyverse/lubridate/issues/314#issuecomment-101816694). I've punted on the first issue (I found `initialize` and `coerce<-` by tedious experimentation), but tried to improve the second one; the PR includes a set of tests which pass for me in all of `devtools::test`, `devtools::check`, `R CMD check --as-cran`, and on Travis.


vspinu:
Ok, so in a nutshell: It's not about Rscript as such. It's simply that when package A has `Depends` on package B,` and package C  `Imports` A, then when package C is loaded, it loads A, but the `Depends` of A, package B is not actually attached.  In our case package C is `methods`.


> I've punted on the first issue (I found initialize and coerce<- by tedious experimentation), 

So I guess, no concrete ideas of how to improve on this?

> but tried to improve the second one; 

Ok, it is good enough. I had something similar before, but removed it for no-idea-why reason at some later stage. Could you push it for CRAN check (aka `devtools::build_win`)? You can leave my email for feedback, I will post it here.

Thanks a bunch for this! Will merge this later today. Could you please add a NEWS entry for this (in bug fixes) and squash all commits in one?

vspinu:
It fails on win-builder https://win-builder.r-project.org/INqPlIrky8xB/00check.log

I recall that last time I wasn't running system checks on CRAN. There is a testhat variable somewhere to inhibit specific tests on CRAN. It's good enough if we could run it on Travis.

infotroph:
Looks like the key line is in [test.that.Rout](https://win-builder.r-project.org/INqPlIrky8xB/examples_and_tests/tests_x64/testthat.Rout.fail): "Fatal error: cannot open file 'R_TESTS=': No such file or directory". I forgot that environment variables and Windows don't go together. 

> It's good enough if we could run it on Travis.

Want me to add `skip_on_cran()`?

infotroph:
Added NEWS entry and skipped tests on CRAN, squashed all commits together.

infotroph:
>> I've punted on the first issue (I found initialize and coerce<- by tedious experimentation),

> So I guess, no concrete ideas of how to improve on this?

No good ideas, no -- I know next to nothing about S4. Unless a wholesale `@import methods` is a workable solution, but I assume if it were that simple you would have done it long ago.

vspinu:
Thanks :+1:

vspinu:
Well, this actually fails locally for me. If it does for me, it will fail for other users. We would need something a bit more robust I guess.

```R
checking tests ... ERROR
  Running ‘testthat.R’
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  do_Rscript("lubridate::dhours(22)") does not match "79200s (~22 hours)".
  Actual values:
  * WARNING: ignoring environment value of R_HOME
  * [1] "79200s (~22 hours)"
  
  
  testthat results ================================================================
  OK: 2367 SKIPPED: 0 FAILED: 4
  1. Failure: methods is not attached (@test-namespace.R#32) 
  2. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#39) 
  3. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#44) 
  4. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#48) 
  
  Error: testthat unit tests failed
  Execution halted
> 

```

```R
1. Failure: methods is not attached (@test-namespace.R#32) ---------------------
do_Rscript("'package:methods' %in% search()") does not match "FALSE".
Actual values:
* WARNING: ignoring environment value of R_HOME
* [1] FALSE


2. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#39) 
do_Rscript("lubridate::round_date(as.POSIXct('2017-10-03 03:01:13Z'), 'minute')") does not match "2017-10-03 03:01:00".
Actual values:
* WARNING: ignoring environment value of R_HOME
* [1] "2017-10-03 03:01:00 CEST"


3. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#44) 
do_Rscript("lubridate::days(1)") does not match "1d 0H 0M 0S".
Actual values:
* WARNING: ignoring environment value of R_HOME
* [1] "1d 0H 0M 0S"


4. Failure: lubridate:: calls work when methods is not attached (@test-namespace.R#48) 
do_Rscript("lubridate::dhours(22)") does not match "79200s (~22 hours)".
Actual values:
* WARNING: ignoring environment value of R_HOME
* [1] "79200s (~22 hours)"
```

vspinu:
Also when trying it directly 

```R
+ . + Warning: running command ''R' CMD install --library=/tmp/RtmpJnwTXb/Rlib --no-docs --no-help --no-demo --no-data --no-test-load /store/Dropbox 2>&1' had status 1
[1] "install: unrecognised option '--library=/tmp/RtmpJnwTXb/Rlib'"
[2] "Try 'install --help' for more information."                   
attr(,"status")
[1] 1
```

vspinu:
I have removed the install part for now. I don't quite follow that part. The assumption is that the built package is already there, right? Does that always hold? What about devtools workflow?

For now I am sticking with the simple version - assume that the right version of the package is already installed and accessible. Seems to be the case on Travis and it's fine for local usage.

infotroph:
1. Thanks! Glad this is even partially useful.

2. Yes, the install part is for devtools -- If I'm following right, devtools::test loads the package namespace on the fly rather than build it on disk, so building it manually seemed like the easiest way to ensure the current version is placed in Rscript's path. But it's slow as well as convoluted, and you could argue fairly that it's trying to reimplement R CMD check inside testthat.

3. The extra output is weeeird... I do not know why expect_match doesn't just accept on the substring:

```
> x=paste(c('* WARNING: blah blah blah', '* [1] "79200s (~22 hours)"'), collapse="\n")
> testthat::expect_match(x, "(~22 hours)", fixed=TRUE)
>
```

I'll poke at this some more and see if I can come up with a more robust approach. Will test it on a wider array of systems this time, and submit it as a fresh PR.

vspinu:
>  why expect_match doesn't just accept on the substring:

The "paste" part is that I have added. System returns multi-line output as a vector of lines. I guess expect_match accepts length one string and picks only first element silently otherwise.

> Will test it on a wider array of systems this time, and submit it as a fresh PR.

Ok. Thanks. 

In any event, I am already pretty happy with this at the moment. Whenever I work I tend to re-install the library anyways, so that extra install is not very useful. If you are fine with that, then I think we are good for now.

PR #587, opened 2017-10-13
Remove '%' from file names

siddharthab:
Please will you consider renaming a file such that it does not have '%' in it.

We have a [build system](https://github.com/grailbio/rules_r) for R packages that does not accept '%' in file names.

I will totally understand if you reject.


vspinu:
I will be happy to fix this as it also creates occasional issues with file and symbol navigation in my editor. 

It will have to wait a bit though as I am in the process of releasing new version to CRAN and it is stuck in review for a week only.

PR #594, opened 2017-10-24
trunc_multi_unit 

muschellij2:
Changed `y <- as.POSIXlt(x)` to `x <- as.POSIXlt(x)` to reduced the memory footprint of `trunc_multi_unit` and overwrite `x` rather than creating a new vector `y`.  Not a large change, but can be useful when `x` is a very long vector (currently using 1b records).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@ Version 1.6.0.9000
 
 ### NEW FEATURES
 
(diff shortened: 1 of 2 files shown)


vspinu:
I doubt this change helps with the problem. GC is unlikely to be happening immediately, so you end up with two copies anyways.  Best solution is to drop it to C and avoid as.POSIXlt altogether. 

I will merge this once I cut the new release. 1.7.0 is pending CRAN's revision at the moment.

PR #616, opened 2017-12-01
Avoid partial match warning on update in round_ceiling when rounding to month

ateucher:
In `round_ceiling` when rounding to month, there is a partial match of `month` to `months` in the `update` call: 

``` r
library(lubridate)
#> 
#> 
#>     date
options(warnPartialMatchArgs = TRUE)
x <- as.POSIXct(c("2009-08-03 12:01:59.23", "2010-08-03 12:01:59.23"), tz = "UTC")
ceiling_date(x, "month")
#> Warning in (function (object, years = integer(), months = integer(), days =
#> integer(), : partial argument match of 'month' to 'months'
#> [1] "2009-09-01 UTC" "2010-09-01 UTC"
```


vspinu:
Thank you!

ateucher:
My pleasure - thanks!

PR #628, opened 2018-01-23
Fixed typo's in parse.r: select_formats argument

schoonees:
Fixed text for the select_formats argument: Se -> See; tockens -> tokens; it -> It


vspinu:
Damn. I guess I was under heavy influence when I wrote that :/  Thanks!

PR #647, opened 2018-03-02
Use lubridate.verbose for all message calls (fixes #173)

### LINKED ISSUE #173: Suppressing "estimate only" message
_I posted this in the lubridate's Google Groups a month ago.  I'm reposting it here, since it looks like it's the only thread in the group._

Is it currently possible to suppress the message "estimate only: convert difftimes to intervals for accuracy" when executing something like the following line?
`as.period(ymd("2010-01-01") - ymd("2009-11-01"))` #From https://github.com/hadley/lubridate/issues/75

Is there a way to suppress this specific message, without suppressing more important warnings that may occur?

If not, would t be helpful if I forked lubridate and modified this method from

```
setMethod("as.period", signature(x = "difftime"), function(x, unit = NULL, ...){
  message("estimate only: convert difftimes to intervals for accuracy")
  seconds_to_period(as.double(x, "secs"))
})
```

to something like

```
setMethod("as.period", signature(x = "difftime"), function(x, unit = NULL, showAccuracyWarning=TRUE, ...){
  if( showAccuracyWarning ) message("estimate only: convert difftimes to intervals for accuracy")
  seconds_to_period(as.double(x, "secs"))
})
```

?

To make the signatures and behavior consistent, I suppose these two lines show be modified as well.
https://github.com/hadley/lubridate/blob/95418730b7beff3508f7044a4ce11326167afc81/R/coercion.r#L370
https://github.com/hadley/lubridate/blob/95418730b7beff3508f7044a4ce11326167afc81/R/coercion.r#L436

I really like the period concept in lubridate.  I deal with human age a lot, such as if kids were eligible for a specific government program during a past event.  The period `10y 0m 1d` is more accurate/appropriate than dividing a difftime of 3653 days by 365.25.  It's even more important when the eligibility threshold is 6 months or 34 months old.

fbreitwieser:


R/intervals.r
```diff
@@ -467,8 +467,11 @@ setMethod("union", signature(x = "Interval", y = "Interval"), function(x, y) {
 
   spans <- as.numeric(ends) - as.numeric(starts)
 
-  if (any(!int_overlaps(int1, int2)))
-    message("Union includes intervening time between intervals.")
+  if (any(!int_overlaps(int1, int2))) {
+    verbose <- getOption("lubridate.verbose")
+    if( isTRUE(verbose) )
```
vspinu: Could simply be `if (getOption("lubridate.verbose", TRUE))`.

R/ops-division.r
```diff
@@ -89,7 +89,10 @@ divisible_period <- function(per, anchor) {
 
 
 divide_period_by_period <- function(per1, per2) {
-  message("estimate only: convert to intervals for accuracy")
+  verbose <- getOption("lubridate.verbose")
+  if( isTRUE(verbose) )
+    message("estimate only: convert to intervals for accuracy")
```
vspinu: I think this message could be removed altogether. All similar messages have been removed in 1.7.0. Plus I think that "convert to intervals" just doesn't make sense in that context.

fbreitwieser: Even better: `if   (isTRUE(getOption("lubridate.verbose", TRUE))`

vspinu: I don't think it's better; that option is logical. Also the default should be FALSE to comply with the usage in other places.

hadley:
Fixes #173

hadley:
@vspinu I finished this off along the lines you suggested; can you please re-review?

vspinu:
Looks good. Thanks! I only wonder if we should use this flag in all those functions currently accepting `quiet` argument.

PR #649, opened 2018-03-05
Fix system timezone memoization.

QuLogic:
The result of `get_system_tz` comes from a `STRING_ELT`, which seems to be automatically garbage collected, causing the static character pointer to point to an invalid location. Duplicating the string is a small leak, but prevents this invalid access.

Alternatively, you could not do the memoization; I'm not sure what the benefit or downside of that is.

This should fix several of the [failures on CRAN](https://cran.r-project.org/web/checks/check_results_lubridate.html).

### DIFF
--- a/src/update.cpp
+++ b/src/update.cpp
@@ -95,7 +95,7 @@ const char* get_system_tz() {
 
 const char* local_tz() {
   // initialize once per session
-  static const char* SYS_TZ = get_system_tz();
+  static const char* SYS_TZ = strdup(get_system_tz());
   const char* tz_env = std::getenv("TZ");
   if (tz_env == NULL) {
     return SYS_TZ;


vspinu:
This makes sense. But I think duplication should be better done within `get_system_tz`. What do you think?

> which seems to be automatically garbage collected

Have you actually seen this with 1.7.3? You should see garbage returned by `lubridate:::C_local_tz()`.  It was indeed the case with 1.7.2 but I changed the memoization in 1.7.3 and haven't noticed the problem since.

> This should fix several of the failures on CRAN.

Not sure. Empty tz which shows there is not "garbage".

QuLogic:
> But I think duplication should be better done within `get_system_tz`. What do you think?

I didn't put it there because there's no guarantee of it only being called once, whereas we know the `static` variable is. Just a slightly lower chance of extra leaking that way.

> Have you actually seen this with 1.7.3? You should see garbage returned by lubridate:::C_local_tz(). It was indeed the case with 1.7.2 but I changed the memoization in 1.7.3 and haven't noticed the problem since.

Yes, I've never used 1.7.2. It works the first time, but not on the second (or maybe third or fourth) call due to the pointers changing around.

>  >   This should fix several of the failures on CRAN.
> Not sure. Empty tz which shows there is not "garbage".

The empty tz is printed out by the outer wrapper; this garbage shows up in some inner frames but is never printed out.

vspinu:
> Just a slightly lower chance of extra leaking that way.

Ok.  Makes sense.

PR #655, opened 2018-03-17
Remove extraneous include.

QuLogic:
Nothing from it is used, plus it's an "internal" cctz file so dropping
the include allows building against an unbundled system cctz.

### DIFF
--- a/src/update.cpp
+++ b/src/update.cpp
@@ -3,7 +3,6 @@
 #include <unordered_map>
 #include "civil_time.h"
 #include "time_zone.h"
-#include "time_zone_if.h"
 #include "utils.h"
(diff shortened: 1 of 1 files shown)


vspinu:
Thanks. It was hanging there from my early experiments.

If not a secret, where do you build against unbundled cctz and what for?

QuLogic:
I am building on Fedora against the [cctz package](https://apps.fedoraproject.org/packages/cctz) I also made. Currently it's only for lubridate and RcppCCTZ, but there are some things outside the R ecosystem that might start using it as well.

There's also a [system version of cctz on Debian](https://packages.debian.org/source/cctz) available.

PR #668, opened 2018-04-20
"date-time", not "date"

wibeasley:
Should POSIXct values like '2012-06-01 12:23' be considered a date-time, not a date?


vspinu:
Indeed.  While colloquial usage of  "date", "date-time" and "time"  is interchangeable, having a clear distinction within the package is a good thing.

PR #677, opened 2018-05-21
Fix typoes

jjchern:
1. Drop parenthesis for `origin` as `origin` is an object, not a function.
2. Add "`" around a piece of code.
3. Add a missing period.


vspinu:
Thanks!

PR #696, opened 2018-07-19
Add Compare method for duration and numeric objects.

### LINKED ISSUE #695: Compare method needed between Duration and numerics for plotting in ggplot2
Plotting Duration objects in `ggplot2` fails because there is no defined `Compare` method between numeric and Duration objects which is required by the underlying continuous scale calculations (reprex and discussion @ tidyverse/ggplot2#2414). 

Is there a good reason why this method hasn't been implemented? I'm happy to submit a PR adding it if deemed appropriate.

dpseidel:
Closes #695. 

Let me know if you prefer to have the coercion go the other way, ie: 

```
setMethod("Compare", signature(e1 = "Duration", e2 = "numeric"),
         function(e1, e2) {
            callGeneric(as.numeric(e1, "secs"), e2)
          })
```

Also will fix tidyverse/ggplot2#2414

### DIFF
--- a/R/durations.r
+++ b/R/durations.r
@@ -352,13 +352,13 @@ setMethod("Compare", c(e1 = "ANY", e2 = "Duration"),
 #' @export
 setMethod("Compare", signature(e1 = "Duration", e2 = "numeric"),
           function(e1, e2) {
-            callGeneric(e1, as.duration(e2))
+            callGeneric(e1@.Data, e2)
           })
 
 #' @export
 setMethod("Compare", signature(e1 = "numeric", e2 = "Duration"),
           function(e1, e2) {
-            callGeneric(as.duration(e1), e2)
+            callGeneric(e1, e2@.Data)
           })
 
 #' @export
--- a/R/durations.r
+++ b/R/durations.r
@@ -349,6 +349,18 @@ setMethod("Compare", c(e1 = "ANY", e2 = "Duration"),
           })
 
 #' @export
+setMethod("Compare", signature(e1 = "Duration", e2 = "numeric"),
+          function(e1, e2) {
+            callGeneric(e1, as.duration(e2))
+          })
+
+#' @export
+setMethod("Compare", signature(e1 = "numeric", e2 = "Duration"),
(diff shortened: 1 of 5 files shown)


vspinu:
Tests need to be  updated. Would be also nice if you could add a few more tests for this specific feature. Thanks!

dpseidel:
Woops sorry about that -- tests are updated and new ones added. Documentation updated. Everything should be passing. 

The existing comparison tests were not exactly were I was expecting so I ended up editing both `test-ops-compare.r` and `test-durations.r`. Let me know if you'd like them consolidated and where. Considering the method is in `durations.r` I would have expected the tests to live in `test-durations.r` but most were already in `test-ops-compare.r` so I added mine there.

R/durations.r
```diff
@@ -348,6 +349,18 @@ setMethod("Compare", c(e1 = "ANY", e2 = "Duration"),
                  call. = FALSE)
           })
 
+#' @export
+setMethod("Compare", signature(e1 = "Duration", e2 = "numeric"),
+          function(e1, e2) {
+            callGeneric(e1, as.duration(e2))
```
vspinu: Sorry to still bug you on this, but I think as.duration is an unnecessary indirection here. You know that comparing to numeric means comparing seconds, so the numeric could be compared directly to e1@.Data as in the "character" method below. 

I even think that an ANY method with as.numeric could be an even better solutions here. But it might have some unexpected consequences though.

dpseidel: No that's fine. Bug away. Let's get it right.  Which do you prefer then? I could try the ANY solution with some tests but I think you're right there may be unforeseen problems cause by that.

dpseidel: For now, I've just changed the method to drop the unnecessary coercion and compare the `@.Data` instead. Makes much more sense.

vspinu: Thanks. Let's do this for now and I will think a bit more about ANY method. Most likely I won't implement it as it will create more room for bugs. For instance, you really don't want it to work on factors.

PR #700, opened 2018-07-26
Add format_ISO8601 with methods

### LINKED ISSUE #629: Feature Request: as.ISO8601
`as.ISO8601` methods would be useful for all the date, date/time, period, duration, and interval functions.

To make this as useful as possible, I think it would benefit to not use `as.character` and make a new `as.ISO8601` generic function with arguments:

* `x`: The object to convert
* `format`: This would depend on what is being represented:
    * With dates, it could be "ymd" (year month day), "ywd" (year week day), "yd" (year day in YYYY-DDD format), "y" (year only), "m" (month only), "md" (month day only)
    * With times, it could be "hms", "hm", or "h"
    * With date/times, it could be any pair of the options for dates and times separated by a "T" (default "ymdThms").  I chose "T" rather than something else to align with the expected output as defined in the ISO standard.
    * Periods would share the format specifier with date, times, or date/times.
* `include_tz`: if `NULL` (default), include/exclude the time zone from the date/time based on the input data.  If `FALSE`, exclude the time zone from the representation, and if `TRUE` include the time zone in the representation.  (It would be ignored if time zone does not apply to the output format.)
* `repeat`: (default: `NULL`) For intervals, repeating could be specified by setting to a non-NULL, positive integer or `Inf` for indefinite, future repeats.
* `repeat_start`: (default: `NULL`) For intervals, repeat start date/time.  (Must be a date, time, or date/time object type.)

(As initially discussed in #362)

billdenney:
Fix #629


R/format_ISO8601.r
```diff
@@ -0,0 +1,137 @@
...
+#'   "ymdhm" would show precision through minutes.
+#' @param max_precision The maximum precision allowed to be output.
+#' @param usetz Include the timezone in the output format
+#' @return A format string ready for \code{as.character} methods.
+#' @details
+#' When \code{NULL}, \code{max_precision} is returned.  When \code{precision} is
+#' more precise than \code{max_precision}, a warning is given and
+#' \code{max_precision} is returned.
+format_ISO8601_precision_check <- function(precision, max_precision, usetz=FALSE) {
+  precision_map <- list(y="%Y",
```
vspinu: This should be stored outside of the function to avoid construction on every call.

R/format_ISO8601.r
```diff
@@ -0,0 +1,137 @@
...
+                        ymdh="%Y-%m-%dT%H",
+                        ymdhm="%Y-%m-%dT%H:%M",
+                        ymdhms="%Y-%m-%dT%H:%M:%S")
+  if (!(max_precision %in% names(precision_map))) {
+    stop("Invalid value for max_precision provided: ", max_precision)
+  }
+  if (is.null(precision)) {
+    precision <- max_precision
+  }
+  if (!(precision %in% names(precision_map))) {
```
vspinu: Micro optimization: `if (is.null(ret <- precision_map[[precision]]) { ...}`

R/format_ISO8601.r
```diff
@@ -0,0 +1,137 @@
...
+    stop("Invalid value for precision provided: ", precision)
+  }
+  if (nchar(precision) > nchar(max_precision)) {
+    warning("More precision requested (", precision, ") than allowed (", max_precision, ") for this format.  Using maximum allowed precision.")
```
vspinu: Could you please split this line ?

R/format_ISO8601.r
```diff
@@ -0,0 +1,137 @@
...
+#'
+#' @param x An object to convert to ISO8601 character format.
+#' @param usetz Include the time zone in the formatting (of outputs including
+#'   time; date outputs never include time zone information).
+#' @param precision The amount of precision to represent with substrings of
+#'   "ymdhms", as "y"ear, "m"onth, "d"ay, "h"our, "m"inute, and "s"econd. (e.g.
+#'   "ymdhm" would show precision through minutes.  When \code{NULL}, full
+#'   precision for the object is shown.
+#' @param ... Additional arguments to methods.
+#' @return A character vector of ISO8601-formatted text.
```
vspinu: Could you please add `@references` with wiki page for ISO8601 format?

vspinu:
You will also have to add S4 aliases, otherwise CRAN check complains.

billdenney:
I think that what I just pushed addresses all of the suggestions (while Travis checks are pending).

vspinu:
Thanks :+1: !

Issue #693, opened 2018-07-09
Conflict between dplyr and lubridate for `union`, `setdiff`, `intersect`


NikNakk:
dplyr and lubridate both redefine the base functions, `union`, `setdiff`, `intersect`. dplyr makes these S3 generics with a special case for `data.frame`. `lubridate` makes them S4 generics with a special case for `Interval`. Depending on the order of package loading, only one of these possibilities will be available. Would it be possible to co-ordinate the approach used, since the packages are often used together?

One thing to note is that executing:

```
library(dplyr)
loadNamespace(lubridate)
```

allows both `union(my_interval, my_interval)` and `union(my_data_frame, my_data_frame)` to work as expected, but if `library(lubridate)` is called, the S3 methods defined in `dplyr` are lost.

Tested using R 3.5.0, lubridate 1.7.4 and dplyr 0.7.6 on Windows 7 and RStudio 1.1.1453.

vspinu:
I don't think moving from S3 to S4 or the other way around is an option for either package. But I wonder if that would help. If two packages define their S3 (or S4) generics for same function won't they overwrite each other anyhow?

I don't see a clean solution for this. Maybe @hadley could suggest something.

In any event namespaced call `lubridate::union` should work as expected.

NikNakk:
I may have missed something, but it strikes me that the most straightforward solution would be to change lubridate to use S3 methods for `Interval`. It would mean that the method would need to check the class of the second parameter, since method dispatch is only based on the first, but otherwise I can't see any big issues. I'd be interested to see what @hadley has to say, since he clearly understands S3/S4 far better than I ever will!

vspinu:
Indeed, that could be in this particular case because we don't have methods where first argument is non-interval. But then still, will there be any inter-operational issues given those two S3 generics? 

This issue will require a bit of digging and, unfortunately, I won't have time to  investigate this in the near future. 

For the record, there are also `setdiff` and `intersect` which we `setGeneric`, which should be changed as well.

vspinu:
Or one can check for existence of the generic and define new one if none has been defined.

jimhester:
Bioconductor defines a intersect generic in BiocGenerics https://github.com/Bioconductor/BiocGenerics/blob/c7ed1f9e8350a65e6d495e53a7d529a14716a46c/R/sets.R#L19-L23.

[...]

PR #717, opened 2018-09-10
Don't export functions created with setGeneric()

hadley:
I'm pretty sure these are not necessary, and are actually harmful because they cause conflicts to be reported where conflicts don't actually exist. I confirmed this by reinstalling then running the following code:

``` r
library(lubridate)
#> 
#> 
#>     date
int1 <- interval(ymd("2001-01-01"), ymd("2002-01-01"))
int2 <- interval(ymd("2001-06-01"), ymd("2002-01-01"))
intersect(int1, int2)
#> [1] 2001-06-01 UTC--2002-01-01 UTC
```


vspinu:
Does this help with #693?

vspinu:
So could this be merged, or we are waiting for Jim to comment on this one?

Fine with me either way. I am not developing for R momentary so please take your time if you need.

hadley:
Let's let this sit for a while.

hadley:
Ok, now I see:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> 
#>     intersect, setdiff, setequal, union
library(lubridate)
#> 
#> 

int1 <- interval(ymd("2001-01-01"), ymd("2002-01-01"))
int2 <- interval(ymd("2001-06-01"), ymd("2002-01-01"))
intersect(int1, int2)
#> [1] 2001-06-01 UTC--2002-01-01 UTC
```

<sup>Created on 2019-11-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

vspinu:
There is a new [revdep failure](https://github.com/tidyverse/lubridate/blob/bf9db3cb1cdbe270319936849de7431a054b5414/revdep/problems.md#xroi) most likely due to this change. Any ideas what is going on? Can this be fixed on our side?

vspinu:
My understanding is that `raster` defines its own generics, and thus overwrites ours when [xROI](https://github.com/bnasr/xROI/blob/master/DESCRIPTION) imports both packages. Looks to mem that we cannot do much here. It should be fixed on xROI or, better, on `raster` package, which should be ported on top of `generics`.

hadley:
Yeah, agreed.

vspinu:
https://github.com/rspatial/raster/issues/107

PR #724, opened 2018-10-20
parse_flex draft

superchordate:
I've been working on a more flexible date ingest function which relies heavily on lubridate, so I figured I'd share it in case you'd like to bring it in.

I use this in loops throw lots of columns at it and convert to date when possible, otherwise leave it unaffected. Critically, it also handles/checks for excel dates, although this is far from perfect hence the ability to turn it off or add min/max definitions.

I tried to run tests but got errors so if you have trouble running them let me know.


vspinu:
Thanks you for the PR. I am afraid this is something specific to your application and it doesn't fit into the overall parsing API of lubridate. Judging from the documentation the purpose of the function is not well defined.  I don't quite see how this is fundamentally different from `ymd_hms(.., truncated = 1)`. What is so special about excel format? There is also a dependency on dplyr which we don't want to have.

Issue #709, opened 2018-08-21
Eliminate stringr dependency?


hadley:
How hard would it be? stringr is now a fairly heavy dependency since it uses stringi.

vspinu:
I treed once but backed up for a forgotten reasons. It's not very hard but not trivial.

vspinu:
Fixed by an awesome contribution from @superchordate. Thanks!

PR #725, opened 2018-10-25
Remove stringr dependency

superchordate:
Hi! I'm a newb at an open source hackathon and submitting this partial fix for issue #709. I found one stringr function str_sub in guess.R and replaced it with a multi-step substring paste. Feels a bit like a hack but a better way didn't come to mind.

If I find time, I'll look into the others - would it be better to submit them all together?

Hope it helps!


vspinu:
Yes. Submitting them all together is better. Thanks!

vspinu:
I didn't mean this to be closed. You can simply add commits if you plan to work on it. If not I will have a look and base my fixes on yours if appropriate.

vspinu:
Travis fails due to the stringr dependency.

R/intervals.r
```diff
@@ -217,7 +217,10 @@ interval <- function(start, end = NULL, tzone = tz(start)) {
 }
 
 parse_interval <- function(x, tz) {
-  mat <- str_split_fixed(x, "/", 2)
+  mat <- matrix(    
+    c( gsub( '(^[^/]+)/(.+$)', '\\1', x ), gsub( '(^[^/]+)/(.+$)', '\\2', x ) ),
```
vspinu: you please add a comment of what this does?

R/parse.r
```diff
@@ -791,7 +791,7 @@ fast_strptime <- function(x, format, tz = "UTC", lt = TRUE, cutoff_2000 = 68L) {
                            "%z"},
                      stop("Unrecognised capture detected; please report this bug"))
 
-      str_sub(fmt, zpos, zpos + attr(zpos, "match.length") - 1) <- repl
+      fmt <- str_sub_lubridate(fmt, zpos, zpos + attr(zpos, "match.length") - 1, repl )
```
vspinu: The convention is to name it with a dot `.str_sub`. No need for lubridate because it's clear enough that it's internal function of lubridate.

R/util.r
```diff
@@ -133,3 +133,17 @@ date_to_posix <- function(date, tz = "UTC") {
...
+  # get the parts of the string to the left and right of the replacement.
+  start.c = substr( x, 1, start - 1 )
+  end.c = substr( x, end + 1, nchar(x) )
+  
+  # paste with replacement in the middle.
+  x <- paste( start.c, replace_with, end.c, sep = "" )
+  
+  return(x)
+
+}
```
vspinu: Need a new line  at the end to avoid various tooling complaining.

DESCRIPTION
```diff
@@ -30,7 +30,6 @@ Depends:
     methods,
     R (>= 3.0.0)
 Imports:
-    stringr,
     Rcpp (>= 0.12.13)
```
vspinu: Need to remove `import("stringr")` from the namespace to avoid check failures.

R/intervals.r
superchordate: Ah, missed one! I've added a comment and removed str_extract, and updated the commit title. Thanks for the feedback.

R/parse.r
superchordate: I changed it anyway while I was in there.

R/util.r
superchordate: Done.

DESCRIPTION
superchordate: Gocha. Done.

superchordate:
I reversed my prior commit and did a force push so the changes are in a single commit. Hopefully, that is the best way - please let me know if you'd prefer a new commit for requested changes. It still says "Changes requested" in red, is there something I need to do to resolve that? Maybe the reverse/force push is the issue?

superchordate:
Oh looks like it doesn't go away until you change the status. Sounds good!

superchordate:
You bet! I'm happy to contribute to such a stellar package that I use daily, and I learned a bit about working with open source so thanks to you for that, that article about commits was really good.

PR #800, opened 2019-10-13
Add seq.interval

wfjvdham:
Will return a sequence of dates starting from the start date of an interval, until the end date of the interval and taking steps defined by the `by` argument.

solves #638


hadley:
This doesn't feel quite right to me because generally you create a `seq()` from one object to another. This is particularly problematic giving the challenge definition of the `seq()` generic, which makes it hard to know whether or not you're obeying the expected contract.

PR #814, opened 2019-11-19
Return integer from year()

### LINKED ISSUE #361: should `year()` return an integer instead of a double?
Is there any advantage/disadvantage to the [`year()`](https://github.com/hadley/lubridate/blob/master/R/accessors-year.r) function returning an integer instead of a numeric?

``` r
> (y <- lubridate::year(as.Date("2012-05-09")))
[1] 2012
> class(y)
[1] "numeric"
```

It's certainly not a real problem, but I've noticed that I almost always wrap it in `as.integer()` because it's (a) a grouping variable in dplyr or (b) being written to a database.

hadley:
I think this is a simple oversight in the original code, and I don't think it should cause problems anywhere since R will almost always convert integer to double as needed. This will give a small performance boost for large vectors since it will avoid a coercion from integer to double.

Fixes #361


vspinu:
I don't think this one is as innocuous as it first meets the eye. S4 dispatch and S4 class membership checks will be affected. Things might break unexpectedly in third party code, but ok. We can deal with this. 

My main issue with the patch is that it's incomplete:

```R
> typeof(year(days(1)))
[1] "double"
> typeof(month(now()))
[1] "double"
```

If we are changing `year` we should change it for all classes and for all other accessors as well. This surely would require the change of the Period's internal structure to integers to avoid the extra conversion. 

So let's sleep on this for a while.

hadley:
I don't think this PR needs to boil the ocean; I can certainly review the other accessors but I think it's fine to improve the current state of the world by returning integers from accessors without rethinking the design of the period objecft.

vspinu:
Let's leave this one till after the release. It's a bit risky. 

Periods should be redesigned at a much lower level than this, so agreed, this is not the time. But other accesors should be tackled for sure on this occasion.

hadley:
Would you be happy with an empirical validation that this is safe? I can change it and then run the reverse dependency checks to see if it breaks anything.

vspinu:
Sure, but with all the relevant accessors fixed if it's not too much trouble for you. 

I can run the revdeps myself once I am done with all your pulls and imediate fixes.

hadley:
I’m travelling at the moment so I might not have the time until later next week.

vspinu:
@hadley, Is this one ready?

hadley:
Yup.

vspinu:
The new [revdeps](revdeps/problems.md) is in. This change broke at least 6 packages, mostly due to tests for `identical`, but there is a S4 class breakage error as well.

vspinu:
Ok, this change even breaks  usages of `dplyr`s own `if_else` like in this code [here](https://github.com/Sayani07/gravitas/blob/4fd9a402c11a93003c898891dfde00cecd35c0bf/R/temp_create_gran.R#L356) 

I think this is a pretty serious breakage. Things might be breaking in production systems for no good reason. I am seriously thinking of reverting this change. @hadley please give me your thoughts.

hadley:
Yeah, technically that breakage is a bug in dplyr which the switch to vctrs will eventually fix. I'd agree that it's safest to revert for now, and we can reconsider in a year or two.

vspinu:
Now that vctrs is there we might consider a lightweight version of lubridate, a rewrite so to say.

msberends:
Hi! Would it be an option to reconsider this implementation, since it's been a while?

As pointed out in #361, it still makes sense that `year()` should return an integer for various reasons 🙂

PR #817, opened 2019-11-19
Ignore tz argument in as_date()

### LINKED ISSUE #674: character method for as_date forces `tz` arg when `format` present, doesn't use it
If you try to convert a string using `lubridate::as_date` while specifying the `format`, you must also set `tz`, but it isn't used.

```r
library(lubridate)
#> 
#> 
#>     date

x <- "05172018"

as_date(x)
#> Warning: All formats failed to parse. No formats found.
#> [1] NA

as_date(x, format = "%m%d%Y")
#> Error in strptime(x, format, tz): invalid 'tz' value

as_date(x, tz = "America/New_York", format = "%m%d%Y")
#> [1] "2018-05-17"

tz(.Last.value)
#> [1] "UTC"
```

Current Github version, 1.7.4.

On the one hand, this is unexpected, surprising behavior, but on the other hand, do timezones _ever_ matter for `Date`'s?

In the very least, seems like you shouldn't need the `tz` arg here; the second option should work.

hadley:
Fixes #674


R/coercion.r
```diff
@@ -670,10 +670,14 @@ setMethod(f = "as_date", signature = "numeric",
 #' @export
 setMethod("as_date", "character",
           function(x, tz = NULL, format = NULL) {
+            if (!missing(tz)) {
```
vspinu: I would prefer `!is.null` here instead. This one will get the warning even if the supplier passes `NULL` as an argument. 

While unlikely in this case, the caller might want to have the same argument contracts as with `as_date_time` on some programatic occasions.

PR #822, opened 2019-11-20
Strip enhances

### LINKED ISSUE #821: Review "enhances" packages
i.e. 'chron', 'fts', 'timeSeries', 'tis'. None of these are tested, I suspect haven't been looked at for a long time so could simply be removed.

hadley:
These packages are not used anywhere in lubridate, so I think it makes sense to drop them.

Fixes #821

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -59,12 +59,8 @@ Suggests:
     testthat
 Enhances:
     chron,
-    fts,
     timeDate,
-    timeSeries,
     tis,
-    tseries,
-    xts,
     zoo
 LinkingTo:
     Rcpp


vspinu:
Does it harm to have those "enhances"? We have a few features for those packages. Exclusively generics I think. Or do you propose to eventually remove that limited support that we provide?

I have never used those packages, but `xts` seems pretty popular.

hadley:
Removing them doesn’t break any tests, so we aren’t testing them and it makes me uncomfortable to claim that we “support” a package without any tests to back that up.

vspinu:
Yeh, I am uncomfortable with not having a defined notion of what we support as well. 

But given that we are bound by backward compatibility shouldn't we move the other way around -- add test for them? It's not hard, a few accessors basically. The problem is that new packages are coming all the time and we cannot support all of them.

Alternatively we could drop the support and narrow the scope of the package, but that might harm a few people badly.

hadley:
I think these packages are listed for historical reasons and I don’t think formally supporting them would be worth it. Removing them just recognises that we are not formally supporting these packages. I don’t think this issue is important enough to spend any more time discussing it, so feel free to close or merge.

vspinu:
>  don’t think this issue is important enough 

Well, to my mind it's all about what path we plan to take in the future 1) narrow to base classes with few exceptions (e.g. zoo) 2) make an effort to offer a good enough generic support for many packages out there

I am fine (n fact in favor of) with strategy 1, but it's by all metrics a BIG backward incompatible change.

I will leave this open for now and will ponder about this some day later.

vspinu:
You are right. Life is too short for that kind of stuff ...

PR #824, opened 2019-11-20
Eliminate ambiguous dispatch for %/%

### LINKED ISSUE #792: Ambiguous S4 method dispatch
``` r

birthday <- ymd("1979-10-14")
(birthday %--% today()) %/% years(1)
#> Note: method with signature 'Timespan#Timespan' chosen for function '%/%',
#>  target signature 'Interval#Period'.
#>  "Interval#ANY", "ANY#Period" would also be valid
#> [1] 39
```

<sup>Created on 2019-09-10 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
Fixes #792

### DIFF
--- a/tests/testthat/test-ops-integer-division.R
+++ b/tests/testthat/test-ops-integer-division.R
@@ -67,3 +67,9 @@ test_that("integer division works with Period denominator (#235)", {
   a <- interval(ymd(20140331), ymd(20140419))
   expect_equal(a %/% months(1), 0)
 })
+
+test_that("interval numerator with period denominator", {
+  int <- interval(ymd("2001-01-01"), ymd("2004-05-01"))
+  expect_warning(out <- int %/% years(1), NA)
+  expect_equal(out, 3)
+})
--- a/tests/testthat/test-ops-integer-division.R
+++ b/tests/testthat/test-ops-integer-division.R
@@ -1,75 +1,25 @@
 context("Integer division operations")
 
 test_that("integer division works for interval numerator", {
-  int1 <- ymd("2010-01-01") %--% ymd("2011-01-01")
-  expect_error(int1 %/% int2)
-  expect_equal(int1 %/% months(1), 12)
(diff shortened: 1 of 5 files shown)


tests/testthat/test-ops-integer-division.R
```diff
@@ -1,69 +1,25 @@
...
-})
-
 
-test_that("integer division works with Period denominator (#235)", {
-  a <- interval(ymd(20140331), ymd(20140519))
-  expect_equal(a %/% months(1), 1)
-  a <- interval(ymd(20140331), ymd(20140419))
-  expect_equal(a %/% months(1), 0)
+  verify_output(test_path("test-ops-integer-division.txt"), {
+    int <- ymd("2010-01-01") %--% ymd("2010-06-01")
```
vspinu: If I understand the change correctly, the new tests constitute a subset of the original tests. Each previous block included a test for recycling. 

Also, call me old school but I don't  quite like this monolitic approach to testing. It's harder to debug, keeps output outside of the test and it's quite more fragile. Small output changes on our or R will cause headaches.

hadley: I think I misgeneralised the intent of the tests after reading a few; you are correct that some are also testing that the recycling rules have been correctly interpreted. I'll add those back in.

But you've misinterpreted the intent of this test; it makes sense as a single complicated because it's easier to see that all 9 cases are covered. The use of `verify_output()` turns it into a regression test, which I think makes more sense when you are testing user facing behaviour, which here is exactly which error message you get (the old use of `expect_error()` is dangerous because any error, even if unexpected, will cause the test to pause). `verify_output()` is designed for testing user output so it doesn't fail on CRAN, making it less fragile and easy to update. For example, if we merge #827 after this PR, it'll be easy to see exactly what's changed.

vspinu:
@hadley I have added recycling and concatenation tests myself. Thanks!

Issue #790, opened 2019-08-29
tz() should error on non-dates


NanisTe:
We unfortunately had lost some time to figure out, that the input of the tz() function is not what is expected, the return of the function is always "UTC" if a not supported type is passed. This is unexpected behaviour and can lead to very strange things further on ones code.

---

Where not able to make reprex working in my machine for this example.

```r
x <- data.frame(datetime = as.POSIXct("2010-01-01 00:00:00", tz = "Europe/Zurich"))
tz (x$datetime)
tz (x["datetime"])
tz (x[["datetime"]])
tz("fajhgoiuwnrlk")
tz(500)
```
Results are
```
> x <- data.frame(datetime = as.POSIXct("2010-01-01 00:00:00", tz = "Europe/Zurich"))
> tz (x$datetime)
[1] "Europe/Zurich"
> tz (x["datetime"])
[1] "UTC"
> tz (x[["datetime"]])
[1] "Europe/Zurich"
> tz("fajhgoiuwnrlk")
[1] "UTC"
> tz(500)
[1] "UTC"
```

hadley:
Fixing this would involve replacing `tz.default()` with `tz.POSIXt()`, `tz.Date()` (for backward compatibility, probably should return `NA` with warning), and new `tz.default()` that displays an informative error message.

hadley:
The challenge with making this change is that many existing lubridate functions assume that `tz()` returns "UTC" for dates and character vectors, so this probably needs a new helper that can be used internally. This makes the problem quite a bit harder than I'd originally thought.

vspinu:
I think this one can be considered taken care of. We now return "UTC" only on Dates, characters and logical NA. Those types are automatically promoted all over the place and "UTC" is the lubridate-wide default time-zone. So, it's a good compromise IMO.

PR #825, opened 2019-11-20
Refactoring time zone accessors

### LINKED ISSUE #675: force_tz() should convert dates to date-times
Perhaps this is intentional, but it strikes me as unexpected:

```r
library(lubridate)
#> 
#> 
#>     date

tz(today())
#> [1] "UTC"

tz(force_tz(today(), "America/New_York"))
#> [1] "UTC"

tz(with_tz(today(), "America/New_York"))
#> [1] "America/New_York"
```

Latest Github version, 1.7.4.

In general, timezones for pure `Date`s seem to be discouraged in that virtually no lubridate functions will ever return a non-UTC date (e.g. `today`, `as_date`, `date`, etc.), but I can't find anything to this effect in the docs.

hadley:
Fixes #675, and makes a start on #790.


R/accessors-tz.r
```diff
@@ -1,92 +1,89 @@
...
-  if (is.character(tzone) && nzchar(tzone))
-    return(tzone)
-  tzone <- attr(as.POSIXlt(x[1]), "tzone")[[1]]
-  if (is.null(tzone))
-    return("UTC")
-  tzone
+tz.POSIXt <- function(x) {
+  tzone <- attr(x, "tzone")
+  if (is.null(tzone)) {
+    ""
```
vspinu: Do you know concrete cases when "tzone" can be null on POSIX objects?

I would like to avoid returning (or using) "" as time zone as much as possible. If you pass "" as tz to base (an lubridate) things behave in system/location dependent way. 

Not sure if we could do better here though.

R/accessors-tz.r
```diff
@@ -1,92 +1,89 @@
...
-tz.zoo <- function(x) {
-  attr(as.POSIXlt(zoo::index(x)), "tzone")[[1]]
+tz.Date <- function(x) {
+  # warning("Dates do not have timezones", call. = FALSE)
+  "UTC"
+}
+
+#' @export
+tz.character <- function(x) {
+  "UTC"
```
vspinu: This will be confusing for character time-stamps that contain zulu indicators. I also want to extend the parser to deal with timezones in the future. Could we just fail here?

R/accessors-tz.r
```diff
@@ -1,92 +1,89 @@
...
+}
+
+#' @export
+tz.logical <- function(x) {
+  if (all(is.na(x))) {
+    "UTC"
```
vspinu: What is the logic behind this corner case? Shouldn't it actually be NA in this case? :)

vspinu:
A few extra thoughts.

 I think it makes most sense to support `tz` only on those objects for which we can do `tz<-`, also the return types of objects should be consistent for `with_tz` and `force_tz`. We cannot convert to POSIXct in one and not in another. 

Also `tz<-` is not exactly an alias because it cannot be used on `Date` but `force_tz` can. But ok, I guess we can  afford to be sloppy here for the sake of better intuition.

R/accessors-tz.r
hadley: The problem is that you can do `hour("2012-03-26")` which calls `as.POSIXlt(x, tz = tz(x))$hour`. I don't think this behaviour is great, but fixing it will require a decent amount of work which I think should probably happen separately.

R/stamp.r
```diff
@@ -165,7 +165,7 @@ stamp <- function(x, orders = lubridate_formats,
   dtm_utc <- force_tz(x, tzone = "UTC")
 
   ## the offset is the duration represented by the difference in time
-  offset_duration = as.duration(dtm_utc - x)
+  offset_duration = difftime(dtm_utc, x)
```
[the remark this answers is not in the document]
hadley: Oops, yes.

R/time-zones.r
```diff
@@ -34,12 +34,12 @@ with_tz <- function (time, tzone = "") {
...
-    if (is.POSIXlt(time)) as.POSIXct(time)
-    else time
-  attr(new, "tzone") <- tzone
-  reclass_date(new, time)
+.with_tz <- function(time, tzone = "") {
+  if (!is.POSIXct(time)) {
+    time <- as.POSIXct(time)
+  }
+  attr(time, "tzone") <- tzone
+  time
```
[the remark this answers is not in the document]
hadley: Resolved using `is.POSIXt()` instead; also added unit tests to ensure that this doesn't get broken again.

R/accessors-tz.r
hadley: Using `"UTC"` here causes this test to fail:

```R
test_that("Updateing on seconds doesn't affect hours", {
  ## https://github.com/tidyverse/lubridate/issues/619

  tt2 <- tt <- Sys.time()
  second(tt2) <- 5
  expect_equal(hour(tt), hour(tt2))
})
```

That's because `Sys.time()` returns a POSIXct without a `tzone` attribute.

R/accessors-tz.r
vspinu: I actually think returning UTC on is not that bad if we assume that those character vectors are passed through our own parsers which all have `tz = "UTC"` by default. We could replace that `as.POSIXlt` with `as_datetime` to make it more concrete.

R/time-zones.r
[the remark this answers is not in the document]
vspinu: Why not do it the other way around? Reclass as before but check explicitly for Date? It would still work on foreign date-time classes this way.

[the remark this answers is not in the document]
hadley: I don’t think it makes sense to set an attribute on a vector unless you know what class it is.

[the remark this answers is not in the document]
vspinu: What I meant is to set it on POSIXct but then reclass to whatever. 

I think we should make `with_tz` generic. It makes sense for intervals for instance.

PR #826, opened 2019-11-20
Ensure empty constructors return 0-length vectors

### LINKED ISSUE #707: Zero-input constructors should return zero-length objects
Rather than an error, as currently:

``` r
library(lubridate)

duration()
#> Error in duration(): No valid values have been passed to 'duration' constructor
period()
#> Error in period(): No valid values have been passed to 'period' constructor
interval()
#> Error in tz(start): argument "start" is missing, with no default
```

This makes more similar in behaviour to base R

```r
integer()
#> integer(0)
factor()
#> factor(0)
#> Levels:
```

And this property is taken advantage of in [vctrs](http://vctrs.r-lib.org)

hadley:
Fixes #707

### DIFF
--- a/R/durations.r
+++ b/R/durations.r
@@ -243,7 +243,9 @@ setMethod("[[<-", signature(x = "Duration"),
 #' @export
 duration <- function(num = NULL, units = "seconds", ...) {
   nums <- list(...)
-  if (is.character(num)) {
+  if (is.null(num) && missing(...)) {
+    .duration_from_num(numeric(), units)
+  } else if (is.character(num)) {
     as.duration(parse_period(num))
   } else if (!is.null(num) && length(nums) > 0) {
     c(.duration_from_num(num, units), .duration_from_units(nums))
--- a/R/durations.r
+++ b/R/durations.r
@@ -92,12 +92,15 @@ compute_estimate <- function (secs, unit = "second") {
 
 #' @export
 setMethod("show", signature(object = "Duration"), function(object) {
-  print(format.Duration(object), quote = TRUE)
+  if (length(object@.Data) == 0) {
+    cat("<Duration[0]>\n")
+  } else {
+    print(format(object), quote = TRUE)
+  }
 })
 
 #' @export
 format.Duration <- function(x, ...) {
-  if (length(x@.Data) == 0) return("Duration(0)")
   out <- vector("character", length(x@.Data))
   nnas <- !is.na(x@.Data)
   out[nnas] <- compute_estimate(abs(x@.Data[nnas]))
--- a/R/durations.r
+++ b/R/durations.r
@@ -175,7 +175,7 @@ setMethod("[[<-", signature(x = "Duration"),
 #'   formats are supported; 'm' stands for month and 'M' for minutes unless ISO
 #'   8601 "P" modifier is present (see examples). Fractional units are
 #'   supported.
-#' @param units a character string that specifies the type of units that num
+#' @param units a character string that specifies the type of units that `num`
 #'   refers to. When `num` is character, this argument is ignored.
 #' @param ... a list of time units to be included in the duration and their
 #'   amounts. Seconds, minutes, hours, days, and weeks are supported.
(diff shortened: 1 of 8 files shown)


R/periods.r
```diff
@@ -393,7 +400,12 @@ setMethod("$<-", signature(x = "Period"), function(x, name, value) {
 #' @export
 period <- function(num = NULL, units = "second", ...) {
   nums <- list(...)
-  if (is.character(num)) {
+  if (is.null(num) && missing(...)) {
```
vspinu: I think this change and the entire spaghetti of the if-elses that follows can be reduced to one if-else with the pre-conversion `num <- as.numeric(num)`. Let me try it out. 

BTW, `period(numeric())` fails ATM.

vspinu:
I have just realized that we have different contracts for legacy duration and period constructors. One is vectorised another is not:

```R
period(c(30, 20), units = c("sec", "days"))
[1] "20d 0H 0M 30S"
> duration(c(30, 20), units = c("sec", "days"))
[1] "30s"                    "1728000s (~2.86 weeks)"
```

vspinu:
@hadley I have simplified the constructors' code and generalized the behavior on empty inputs. Could you please check if you are ok with it?

R/durations.r
```diff
@@ -242,28 +245,25 @@ setMethod("[[<-", signature(x = "Duration"),
...
+    return(new("Duration", numeric()))
+
   if (!is.numeric(num)) {
     stop(sprintf("First argument to `duration` constructor must be character or numeric. Supplied object of class '%s'", class(num)))
   }
 
   ## qucik check for common wrongdoings: https://github.com/hadley/lubridate/issues/462
   if (class(num)[[1]] %in% c("Interval", "Period"))
-    stop("Interval or Period objects cannot be used as input to 'period' constructor. Plese use 'as.duration'.")
+    stop("Interval or Period objects cannot be used as input to 'duration' constructor. Use 'as.duration' instead.")
```
hadley: If you're touching this, I think you should update to:

```suggestion
    stop("Interval or Period objects cannot be used as input to `duration()` constructor. Use `as.duration()` instead.", call. = FALSE)
```

R/periods.r
```diff
@@ -418,13 +421,15 @@ parse_period <- function(x) {
...
+    return(new("Period", numeric()))
 
   if (!is.numeric(num)) {
     stop(sprintf("First argument to `period` constructor must be character or numeric. Supplied object of class '%s'", class(num)))
   }
 
-  if (is.interval(num))
-    stop("Interval objects cannot be used as input to 'period' constructor. Plese use 'as.period'.")
+  if (class(num)[[1]] %in% c("Interval", "Duration"))
+    stop("Interval or Durations objects cannot be used as input to 'period' constructor. Plese use 'as.period'.")
```
hadley: Should be consistent with error message above.

R/periods.r
```diff
@@ -418,13 +421,15 @@ parse_period <- function(x) {
...
+  if (length(num) == 0)
+    return(new("Period", numeric()))
 
   if (!is.numeric(num)) {
     stop(sprintf("First argument to `period` constructor must be character or numeric. Supplied object of class '%s'", class(num)))
   }
 
-  if (is.interval(num))
-    stop("Interval objects cannot be used as input to 'period' constructor. Plese use 'as.period'.")
+  if (class(num)[[1]] %in% c("Interval", "Duration"))
```
hadley: This will fail for subclasses of Interval and Duration, which is probably not what you want.

vspinu: yes, this is what we use elsewhere for efficiency reasons (I guess). Let me change all of that.

vspinu:
>  I think a merge is a bit better when working on a PR branch because it makes it a bit easier for the other contributor (particularly newer github users) to pull back the changes.

Yeh, it's a trade-off between clean history and ease of use. Good that you are not the "newer github user" ;)

hadley:
We now generally do a "squash and merge" when merging, since maintaining the history within a PR isn't usually that important.

PR #827, opened 2019-11-20
Binary errors

hadley:
Some tests in #824, but when I started looking into tests I discovered a lot of ambiguous S4 dispatch that will also need to be resolved.

### DIFF
--- a/R/durations.r
+++ b/R/durations.r
@@ -339,11 +339,8 @@ summary.Duration <- function(object, ...) {
 
 #' @export
 setMethod("Compare", c(e1 = "Duration", e2 = "ANY"),
-          function(e1, e2) {
-            stop(sprintf("Incompatible duration classes (%s, %s). Please coerce with `as.duration()`.",
-                         class(e1), class(e2)),
-                 call. = FALSE)
-          })
+  function(e1, e2) stop_incompatible_classes(e1, e2, .Generic)
+)
 
 #' @export
 setMethod("Compare", c(e1 = "ANY", e2 = "Duration"),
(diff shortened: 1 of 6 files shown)


NAMESPACE
```diff
@@ -250,6 +250,7 @@ exportMethods(rep)
 exportMethods(setdiff)
 exportMethods(show)
 exportMethods(union)
+import(rlang)
```
vspinu: Extra dependency for the sake of `abort`?

vspinu:
I am kinda divided on this one. Previous errors are more explicit, but well, the new parsimony might well be worth it.

NAMESPACE
hadley: Yeah, just because we’re likely to use it for other tools eventually too. rlang has no other dependencies so we generally consider it a “free” dependency for tidyverse packages.

hadley:
They were more explicit but I’m not sure if they’re more helpful — for example, it’s not clear to me that they actually give consistent advice, especially in the presence of ambiguous dispatch.

vspinu:
Ok, let's go with this then. I think this ambiguous dispatch is a new thing. I don't recall dealing or even noticing it before.

vspinu:
@hadley I have merged this but without the rlang part for now. Let's pull it in once we have a clear use case leveraging rlang. Abort vs stop is not enough IMO.

hadley:
I'm not sure what you've done with the force-pushing, but now you can't see what this PR contained, and what you vs. I contributed.

vspinu:
I have cherry picked the relevant comit, amended and deleted the tangling branch. By looking at the history in the master it should be clear what is what, but I guess you have figured it by now. Should have waited with the deletion a bit longer.

PR #828, opened 2019-11-22
addition works with infinite dates and timespans

### LINKED ISSUE #732: Mathematical operations with infinite dates and datetimes
Philosophical considerations aside, I think that last one should be TRUE as well.

```
library(lubridate)
infinity <- as_date(Inf)
is.infinite(infinity)
# TRUE
is.infinite(infinity + make_difftime(12, "hours"))
# TRUE
is.infinite(infinity + hours(12))
# FALSE
```

keesdeschepper:
Fixes #732 . Actually two bugs. For `add_duration_to_date` it had to be tackled that hour(infinity) etc. produces `NA`. For `add_period_to_date` the problem was that the date is tranformed to a POSIXlt for the mathematics, which is a class that does not account for infinity. The POSIXlt part is left as it was -- only after the answer is transformed back to the original date class there is a correction for infinite outcomes.

### DIFF
--- a/R/ops-addition.r
+++ b/R/ops-addition.r
@@ -17,11 +17,10 @@ add_duration_to_date <- function(dur, date) {
     date <- as.POSIXct(date)
     ans <- with_tz(date + dur@.Data, "UTC")
 
-    if (all(is.na(ans))) return(as.Date(ans))  # ALL NAs
-
-    if (all(hour(na.omit(ans)) == 0  &
-            minute(na.omit(ans)) == 0 &
-            second(na.omit(ans)) == 0)) {
+    if (all(is.na(ans) |
+            is.infinite(ans) |
+            is.nan(ans) |
+            (hour(ans) == 0 & minute(ans) == 0 & second(ans) == 0))) {
       return(as.Date(ans))
     }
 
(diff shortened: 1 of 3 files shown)


R/ops-addition.r
```diff
@@ -17,11 +17,10 @@ add_duration_to_date <- function(dur, date) {
     date <- as.POSIXct(date)
     ans <- with_tz(date + dur@.Data, "UTC")
 
-    if (all(is.na(ans))) return(as.Date(ans))  # ALL NAs
-
-    if (all(hour(na.omit(ans)) == 0  &
-            minute(na.omit(ans)) == 0 &
-            second(na.omit(ans)) == 0)) {
+    if (all(is.na(ans) |
```
hadley: This logic doesn't seem quite right to me — I think it would be better to decide to return a date or not depending on whether the duration is a integer multiple of 86400 or not. Then I think this logic would get a lot simpler.

R/ops-addition.r
```diff
@@ -58,7 +57,11 @@ add_period_to_date <- function(per, date) {
   if (is.Date(date) && sum(new$sec, new$min, new$hour, na.rm = TRUE) != 0)
```
hadley: As above, I think it would be better to decide on whether or not to return a date based on the values of the period. Do you want to have a go at making this change?

keesdeschepper: I do.

keesdeschepper: Just to be clear: so `as_date(NA) + seconds(1))` should also be a datetime, not a date. This is broader than infinites. New Issue?

hadley: Right

vspinu:
Timechange is finally in and the referenced issue is fixed.

PR #832, opened 2019-11-26
Changing hour, minute or second of Date object now yields POSIXct.

### LINKED ISSUE #831: changing hour, minute or second of date should yield POSIXct
See https://github.com/tidyverse/lubridate/issues/732#issuecomment-558321821.
Changing hour, minute or second of date now yields POSIXlt:
```r
class(as_date(0) + hours(1))
#> "POSIXlt" "POSIXt" 
```
POSIXct is the preferred class for datetimes, though (cf. `as_datetime`). Moreover, POSIXlt does not allow for infinity, leading to bugs like: 
```r
as.numeric(as_date(c(0, Inf)))
#> 0 Inf
as.numeric(as_date(c(0, Inf)) + hours(1))
#> 3600   NA
```

keesdeschepper:
Fixes #831 .

### DIFF
--- a/R/ops-addition.r
+++ b/R/ops-addition.r
@@ -49,14 +49,20 @@ add_period_to_date <- function(per, date) {
   ms <- month(per) + year(per) * 12
   lt <- add_months(lt, ms)
 
+  if (is.Date(date)) {
+    new <- update(as.Date(lt),
+                  days = mday(lt) + per@day,
+                  hours = per@hour,
+                  minutes = per@minute,
+                  seconds = per@.Data)
+    return(new)
+  }
+
   new <- update(lt,
                 days = mday(lt) + per@day,
                 hours = hour(lt) + per@hour,
                 minutes = minute(lt) + per@minute,
                 seconds = second(lt) + per@.Data)
-
-  if (is.Date(date) && sum(new$sec, new$min, new$hour, na.rm = TRUE) != 0)
-    return(new)
 
   reclass_date(new, date)
 }
--- a/tests/testthat/test-settors.R
+++ b/tests/testthat/test-settors.R
@@ -95,30 +95,37 @@ test_that("seconds settor returns NA for spring dst gap", {
 
 })
 
-test_that("seconds settor retains object class", {
+test_that("seconds settor retains object class for datetimes", {
   poslt <- as.POSIXlt("2010-02-14 01:59:59", tz = "GMT", format
      = "%Y-%m-%d %H:%M:%S")
   posct <- as.POSIXct(poslt)
-  date <- as.Date(poslt)
 
   second(poslt) <- 10
   second(posct) <- 10
-  day(date) <- 10
 
-  expect_is(poslt, "POSIXlt")
-  expect_is(posct, "POSIXct")
-  expect_is(date, "Date")
+  expect_s3_class(poslt, "POSIXlt")
+  expect_s3_class(posct, "POSIXct")
 
   second(poslt) <- 70
   second(posct) <- 70
-  day(date) <- 32
 
-  expect_is(poslt, "POSIXlt")
-  expect_is(posct, "POSIXct")
-  expect_is(date, "Date")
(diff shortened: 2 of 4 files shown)


tests/testthat/test-settors.R
```diff
@@ -95,30 +95,37 @@ test_that("seconds settor returns NA for spring dst gap", {
...
-  expect_that(date, is_a("Date"))
 
 })
 
+test_that("seconds settor makes POSTXct from Date", {
+  date <- as.Date("2010-02-14", tz = "GMT", format = "%Y-%m-%d")
+
+  second(date) <- 10
+
+  expect_that(date, is_a("POSIXct"))
```
hadley: Can you please update to use modern style, i.e. `expect_s3_class(date, "POSIXct")`

R/ops-addition.r
```diff
@@ -56,7 +56,7 @@ add_period_to_date <- function(per, date) {
                 seconds = second(lt) + per@.Data)
 
   if (is.Date(date) && sum(new$sec, new$min, new$hour, na.rm = TRUE) != 0)
-    return(new)
+    return(as.POSIXct(new))
```
hadley: Is it possible to fix this upstream? (i.e. in `update()`)?

keesdeschepper: You mean let update.Date always return a POSIXct? Well, the update here is update.POSIXlt preceded by add_months. So also rewrite add_months to have Date output?

hadley: Ok, if that's going to involve going down a rabbit hole, it's fine to fix here.

tests/testthat/test-settors.R
hadley: Sorry, I meant _just_ for the code that you touched. Do you mind undoing the changes above (to make it easier to remove), and only using `expect_s3_class()` in the tests that you have written?

R/ops-addition.r
```diff
@@ -43,33 +43,28 @@ add_period_to_period <- function(per2, per1) {
 
 add_period_to_date <- function(per, date) {
 
-  lt <- as.POSIXlt(date)
+  new <- update(date,
```
vspinu: You have replaced 1 update call with 3 and introduced quite a number of POSIXct -> POSIXlt conversions by dropping the original `lt <- ` line. This will surely have a dire consequence on speed.

Why do you need so many changes if  you can just check if output is divisible by 86400 as @hadley proposed in the other thread?

R/ops-addition.r
```diff
@@ -43,33 +43,28 @@ add_period_to_period <- function(per2, per1) {
...
-                seconds = second(lt) + per@.Data)
+  new <- update(new,
+                days = mday(date) + per@day,
+                hours = hour(date) + per@hour,
+                minutes = minute(date) + per@minute,
+                seconds = second(date) + per@.Data)
 
-  if (is.Date(date) && sum(new$sec, new$min, new$hour, na.rm = TRUE) != 0)
-    return(new)
+  if (is.Date(date)) return(new)
```
vspinu: So you are always returning datetime now?

keesdeschepper: Thanks for the feedback! The rationale was that update.Date already has a check on zero h/m/s (also sufering from the NA issue so should be rewritten as well). By avoiding the conversion to POSIXlt, update.Date would be used for Dates instead of update.POSIXlt. But you are a right about the performance. There is the option to combine both worlds though. Keep the conversion to POSIXlt with the years/months addition, then have an if for Dates that converts back to a Date and calls update.Date. Microbenchmark showed this was faster for Dates then the current implementation. What do you think?

vspinu: Yes, that's a good idea. I think you can even avoid the conversion back to date if the if-else is big enough with an update on the `date` argument directly, but let's not bother with stylistics for now. All of these code will go away once we move on top of time-change. 

Please note that I might not merge this PR before the next release (very urgent) as people might rely on POSIXlt output and I don't want to break stuff right now.

PR #838, opened 2019-11-29
Remove issue template.

### LINKED ISSUE #837: Get rid of intermediate page for making issues
@vspinu, I'm turning this into an issue (which, of course, meant I went through the intermediate page, ironically) to keep better track of it!

Per comment [here](https://github.com/tidyverse/lubridate/commit/208c8700e00cf651c99ce3d743f463bf9a60af0f#commitcomment-36184403)
> Do you know how to disable intermediate question-page with one option which creps between the user pressing [New Issue] button and the actual issue page? I find this intermediate page useless and slightly annoying. Thanks.

Goal is to get rid of the page that automatically appears when you go to create a new issue (see below).

image

Currently, you can create a blank issue by clicking on the <ins>Open a blank issue</ins> in the interface above, but you do need to go through that step.

Temporary fix would be to just get rid of the issue template (introduced in https://github.com/tidyverse/lubridate/commit/208c8700e00cf651c99ce3d743f463bf9a60af0f) entirely. 

Looking at GitHub's ["Configuring issue templates for your repository"](https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository) and ["About issue and pull request templates"](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates), it doesn't seem like there's currently a way to have the issue templates available without forcing the choosing interface.

batpigandme:
* Fixes #837


vspinu:
Let's not remove it, getting rid of the folder will most likely do the trick. 

But, I would rather have the template less verbose, but that's a different issue.

batpigandme:
OK, I moved the issue template up to the `.github/` directory.

vspinu:
Ok. Let's see if it works. Thanks!

PR #841, opened 2019-12-05
Separate *! markers from date-time formats

DesiQuintans:
I was briefly confused when  `d` was presented in the `parse_date_time()` docs as `d!`, and wondered whether there had been a UI change where this `!` was meaningful. My change visually separates these markers from the date-time format tag so that it is clearer that they are not part of the tag itself.


hajebn:
Thanks a lot and good luck, God willing

vspinu:
Thanks! Much better indeed.

Issue #855, opened 2020-01-31
`make_date` and `make_datetime` not consistant for out of range values


brunj7:
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.

Please include a minimal reproducible example (AKA a reprex). If you've never heard of a [reprex](http://reprex.tidyverse.org/) before, start by reading <https://www.tidyverse.org/help/#reprex>.

related to #839

`make_date` and `make_datetime` are not consistent when out of range values are passed as parameters. `make_date` will generate NA, `make_datetime` will rollover silently

```r
``` r
library(lubridate)
#> 
#> Attaching package: 'lubridate'
#> The following object is masked from 'package:base':
#> 
#>     date
make_datetime(2013, 2, 40)
#> [1] "2013-03-12 UTC"
make_date(2013, 2, 40)
#> [1] NA
```

<sup>Created on 2020-01-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>
```

Kodiologist:
Worse than the fact that they're inconsistent is that both are fail-silent.

DanChaltiel:
This is actually a very useful feature to have a rollover, but it is surprising at first and you cannot disable it.
IMHO, there should be a `rollover=TRUE` parameter to opt-in in both functions.

PR #856, opened 2020-02-01
Update documentation as-date

### LINKED ISSUE #839: Feature request: fail-noisy equivalents of `make_date`, `as_datetime`, etc.
Functions that return `NA` for invalid input like `2007-13-01` or `2007-02-29`, rather than raising an error, and don't even print a warning, feel dangerous to me. I need to remember to check for `NA`s every time or I could create a hard-to-find bug.

brunj7:
Fixed #839

The original request was tackling 2 problems
- There was no mention in the documentation of  NA generation by `as_date` and `as_datetime`
- Generate a warning for `make_date` and `make_datetime`, but I found the 2 functions were not consistent how they handle out of range value (see #855) => not implemented in this pull request

### DIFF
--- a/R/coercion.r
+++ b/R/coercion.r
@@ -604,6 +604,7 @@ setMethod("as.character", signature(x = "Interval"), function(x, ...) {
 #'      a more intuitive conversion (see examples)
 #'   \item Both functions provide a default origin argument for numeric
 #'      vectors.
+#'   \item Both functions will generate NAs for invalid date format. A warning message will provide a count of the elements that were not converted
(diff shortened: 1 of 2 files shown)


krlmlr:
@hadley: This looks good to me. Solves part of the original issue, the new issue #855 covers the rest.

Kodiologist:
> Both functions will generate NAs for invalid date format

I think you mean something like "invalid character input", since `as_date("2013-02-29")` also returns `NA`, even though the format is correctly recognized.

vspinu:
Thanks!

Issue #810, opened 2019-11-18
Update testthat style


hadley:
From `expect_that()` to modern form. Need to split into multiple small issues (one per file? one per set of files?) for tidyverse developer day

hadley:
``` r
library(tidyverse)
library(fs)

setwd("~/Documents/tidyverse/lubridate/")

dir_ls("tests/testthat/", glob = "test*") %>% 
  map(read_lines) %>% 
  map_int(~ sum(str_detect(.x, "expect_that"))) %>% 
  enframe() %>% 
  filter(value > 0)
#> # A tibble: 13 x 2
#>    name                                     value
#>    <chr>                                    <int>
#>  1 tests/testthat/test-accessors.R             94
#>  2 tests/testthat/test-decimal-date.R           1
#>  3 tests/testthat/test-durations.R             12
#>  4 tests/testthat/test-ops-addition.R          30
#>  5 tests/testthat/test-ops-division.R          14
#>  6 tests/testthat/test-ops-multiplication.R    12
#>  7 tests/testthat/test-ops-subtraction.R       10
#>  8 tests/testthat/test-parsers.R              187
#>  9 tests/testthat/test-periods.R               31
#> 10 tests/testthat/test-settors.R              315
#> 11 tests/testthat/test-timespans.R             19
#> 12 tests/testthat/test-timezones.R             10
#> 13 tests/testthat/test-update.R               180
```

<sup>Created on 2019-11-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
`test-settors` needs to be split up to have one file per R file.  `test-parsers` and `test-updates` would benefit from refactoring to use more helper functions, but that's just as easy to do after switching to the new form.

sushmitavgopalan16:
@hadley 
to split up test-settors into separate files, does this logic seem ok? 

# Splitting test-settors.R into separate files 

## test-seconds-settor.R

- seconds settor correctly performs simple updates
- seconds settor rolls over as expected
- seconds settor does not change time zone
- seconds settor returns NA for spring dst gap
- seconds settor retains object class

## test-minutes-settor.R

- minutes settor correctly performs simple updates
- minutes settor rolls over as expected
- minutes settor does not change time zone
- minutes settor returns NA for spring dst gap
- minutes settor retains object class

## test-hours-settor.R

- hours settor correctly performs simple updates
- hours settor rolls over as expected
- hours settor does not change time zone
[...]

Issue #819, opened 2019-11-19
Restyle files with stlyer


hadley:
To ensure consistent with latest tidyverse styling guidelines; this makes it easier for others contribute.

hadley:
Needs to wait until no outstanding PRs

lorenzwalthert:
What about using https://lorenzwalthert.github.io/precommit/? I aim for a CRAN release before the end of the year. I am also happy to transfer the repo to r-lib if this would be a pre-condition to use it in any tidyverse repo.

hadley:
I don’t really like the pre-commit approach, and if we did use it, it would be better to pilot it on a package that is already close to the standard style.

lorenzwalthert:
Ok. May I ask why you don’t like it?

hadley:
It requires conda so that you can install https://pre-commit.com, which is a "multi-language package manager". It feels like you are solving a small problem with a large chain saw. I'd much rather solve with github actions.

lorenzwalthert:
Ok, thanks for clarifying. You need at least python for pre-commit.com, the R package provides a default installation mechanism easily accessible from R that requires conda, but the executable can also be installed differently and later be linked to use the R package. 
On the other hand, my understanding of using styler and other *correction tools* with ghactions is that it will create new commits and convolute the git history whereas the pre-commit approach can tackle problem before they enter version control. Need to learn more about ghactions though.

(sorry to lead this discussion off topic)

vspinu:
> To ensure consistent with latest tidyverse styling guidelines; 

I hope they are not too opinionated and Rstudio centric. Otherwise, I am completely fine with this idea.

PR #859, opened 2020-02-12
update more tests to modern style

sushmitavgopalan16:
partially addresses #810


sushmitavgopalan16:
i'm not sure if i should be running styler based on https://github.com/tidyverse/lubridate/issues/819, so i'll reverse it on the two files i ran it on

hadley:
Thanks! It's slightly better to not restyle (saving it to do one restyling pass at the end), but it's not bad enough to make it worth undoing. Thanks for continuing to work on these tests!

PR #862, opened 2020-02-18
Split test settors

sushmitavgopalan16:


vspinu:
Ok, this breakdown seems more meaningful than in the original proposal. Do you plan to move `getters` here as well? So far there are only setters tested here, right?

vspinu:
@sushmitavgopalan16 ping :)

sushmitavgopalan16:
@vspinu @hadley apologies! i went off vacation with just an ipad - will be back on sunday and follow up on this asap!

PR #924, opened 2020-10-11
rename check_ymd to check_mdays; remove unused argument

TianyiShi2001:
This function actually checks if the day number is valid in a given month, so it's better named `check_mdays`. Also, there is one unused argument, which I've removed.


vspinu:
Thanks!

PR #934, opened 2020-11-12
Fix tm_gmtoff build failure on AIX

kadler:
Port of 98fb541 from upstream cctz to fix builds on AIX, see
https://github.com/google/cctz/pull/177


vspinu:
Perfect. Thanks!

PR #938, opened 2020-12-08
Prefer R_alloc to declaring variable-size array

MichaelChirico:
IINM the variable-size array is vulnerable to stack overflow segfault.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -24,6 +24,7 @@
 #include "utils.h"
 #include <stdio.h>
 #include <ctype.h>
+#include <Rinternals.h>
 
 // return adjustment (in seconds) due to leap years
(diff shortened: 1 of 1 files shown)


vspinu:
Did it blow for you?

That's a boolen marker. Let's move it to bool, or even bitwise. WDYT?

R_alloc can be for sure.

MichaelChirico:
It didn't bite no... just happened to notice when I was in the source for some reason.

I'm not sure what you mean by bitwise (my first reaction is to store it like flags: `000100010110` but that would require a potentially huge storage type so I guess that's not it).

Going to push a move to `bool`.

vspinu:
Thanks!

PR #939, opened 2020-12-10
Documentation: 19th and 20th century ambiguity

slodge:
This is very minor, but is a small proposed change to the wording for `cutoff_2000`

The 19th/20th centuries mentioned in the current comment feels misleading. Technically (and we all <3 being technical) the 21st century started 1st Jan 2001 - see https://en.wikipedia.org/wiki/21st_century 

To avoid confusion/misinterpretation I've changed this to simply talk about "starting with `19`" and "starting with `20`"


vspinu:
Thanks! It was a typo actually. Of course it's not 19th century :)

PR #946, opened 2021-01-18
Refactor tzdir_set() into own function

hadley:
I added a warning to make it more when somthing has gone wrong, and removed the special case behaviour for the cctz default directory to make TZDIR more precise.

If we keep seeing tzdir problems, I think we'll need to extract out an exported helper that helps us diagnose the problem.


vspinu:
Any chance this could be reverted in R? The current approach doesn't look right at all. I can open an issue, unless it has been discussed already somewhere.

vspinu (approved):

hadley:
@vspinu you can try asking on R-devel, but (a) given the people involved you will have a hard time persuading them to change and (b) it's already out in the wild so we'd still need to keep this code anyway.

PR #966, opened 2021-04-05
Switch from Rcpp to cpp11

DavisVaughan:
FYI @hadley 

This is a fairly straightforward translation from Rcpp to cpp11. The main thing to note is that cpp11 is less lenient with automatic conversions, so there are `cpp_*()` wrapper functions that enforce certain atomic types before passing on to C++.

@vspinu if you have time to review this, that would be great! Each commit should be very self-contained.

f793787 this commit also fixes a small bug in `get_current_tz()`. If you set `Sys.setenv(TZ = "")` and then call `C_local_tz()`, it would error since `tz_from_R_tzone()` expects a tzone attribute, not a POSIXlt object.


hadley (approved):
Looks great!

src/update.cpp
```diff
@@ -245,7 +259,8 @@ Rcpp::newDatetimeVector C_update_dt(const Rcpp::NumericVector& dt,
   }
   load_tz_or_fail(tzto, tzone2, "CCTZ: Unrecognized tzone: \"%s\"");
 
-  Rcpp::NumericVector out(N);
+  cpp11::writable::doubles out(N);
```
vspinu, line 262: Why do you use `writable` on output vectors?  Can you even return cpp11::doubles back to R? Is there a performance penalty in writable constructor?

vspinu (commented):
Guys, this is amazing. I somehow managed to miss the cpp11 initiative. My heart is singing. Thanks for all the work!

src/update.cpp
DavisVaughan: > Can you even return cpp11::doubles back to R

I'm fairly certain that you can, but you can't modify them at the C++ level using the cpp11 API

> Is there a performance penalty in writable constructor?

No, AFAIK it just exposes the ability write to elements with syntax like `out[i] = value`. The constructor itself essentially just eventually calls this reserve method that calls `Rf_allocVector()` and then shoves the allocated memory in a doubly linked list object that can preserve it
https://github.com/r-lib/cpp11/blob/5892862b77b59e4a443e4c58320e5496530e9a30/inst/include/cpp11/doubles.hpp#L104-L114

A few things you may find interesting
https://cpp11.r-lib.org/articles/converting.html#class-comparison-table
https://cpp11.r-lib.org/articles/converting.html#read-only-vs-writable-vectors-1
https://cpp11.r-lib.org/articles/converting.html#fewer-implicit-conversions-1

The process I went through to convert lubridate over to cpp11
https://cpp11.r-lib.org/articles/converting.html#mechanics-of-converting-a-package-from-rcpp-1

vspinu: Ok, I got it.

PR #986, opened 2021-08-18
emphasize the difference between list and vector for second arg of `%within%` function

yfarjoun:


R/intervals.r
```diff
@@ -544,7 +544,7 @@ setdiff.Interval <- function(x, y, ...) {
 #' @param a An interval or date-time object.
 #' @param b Either an interval vector, or a list of intervals.
 #'
-#'   If `b` is an interval it is recycled to the same length as `a`.
+#'   If `b` is an interval (or interval vector) it is recycled to the same length as `a`.
```
vspinu, line 547: Generally, in R everything is a vector, so when you say x is a double, it is assumed it's a vector, but ok. I guess with intervals, it's not that obvious indeed.

vspinu (approved):

vspinu:
Thanks!

PR #995, opened 2021-09-21
Disambiguate `<Duration> - <Duration>` and `<Period> - <Period>`

### LINKED ISSUE #994: Disambiguate S4 dispatch with subtraction and two Period/Duration objects
We currently get a once per session warning with subtraction when you have two Period or two Duration objects. This is because we have a method for `<ANY> - <Period>` and `<Period> - <ANY>` and it doesn't know which to choose.

``` r
library(lubridate)
#> 
#> 

x <- years(4)
y <- dyears(4)

x - x
#> Note: method with signature 'Period#ANY' chosen for function '-',
#>  target signature 'Period#Period'.
#>  "ANY#Period" would also be valid
#> [1] "0S"

x - x
#> [1] "0S"

y - y
#> Note: method with signature 'Duration#ANY' chosen for function '-',
#>  target signature 'Duration#Duration'.
#>  "ANY#Duration" would also be valid
#> [1] "0s"

y - y
#> [1] "0s"
```

<sup>Created on 2021-09-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

DavisVaughan:
Closes #994 

This PR adds two `-` S4 methods to disambiguate the case where you subtract two Period or two Duration objects.

This is hard to test because the Note appears once per session, but I've added a comment to try and keep us from accidentally removing this in the future.

### DIFF
--- a/R/periods.r
+++ b/R/periods.r
@@ -113,28 +113,29 @@ setClass("Period", contains = c("Timespan", "numeric"),
 #'   /,Period,Period-method /,Period,difftime-method /,difftime,Period-method
 #'   /,Period,numeric-method /,numeric,Period-method *,Period,ANY-method
 #'   *,ANY,Period-method -,Period,ANY-method -,Period,missing-method
-#'   -,ANY,Period-method %%,Period,Duration-method %%,Period,Interval-method
-#'   %%,Period,Period-method >,Period,Period-method >=,Period,Period-method
-#'   ==,Period,Period-method !=,Period,Period-method <=,Period,Period-method
-#'   <,Period,Period-method >,Period,Duration-method >=,Period,Duration-method
-#'   ==,Period,Duration-method !=,Period,Duration-method
-#'   <=,Period,Duration-method <,Period,Duration-method >,Duration,Period-method
-#'   >=,Duration,Period-method ==,Duration,Period-method
-#'   !=,Duration,Period-method <=,Duration,Period-method
-#'   <,Duration,Period-method >,Period,numeric-method >=,Period,numeric-method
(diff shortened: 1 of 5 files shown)


R/periods.r
```diff
@@ -113,28 +113,29 @@ setClass("Period", contains = c("Timespan", "numeric"),
...
-#'   !=,Period,Period !=,Period,numeric !=,numeric,Period %%,Period,Duration
-#'   %%,Period,Interval %%,Period,Period *,ANY,Period *,Period,ANY -,ANY,Period
-#'   -,Period,Interval -,Period,missing /,numeric,Period <,Duration,Period
-#'   <,Period,Duration <,Period,Period <,Period,numeric <,numeric,Period
-#'   <=,Duration,Period <=,Period,Duration <=,Period,Period <=,Period,numeric
-#'   <=,numeric,Period ==,Duration,Period ==,Period,Duration ==,Period,Period
-#'   ==,Period,numeric ==,numeric,Period >,Duration,Period >,Period,Duration
-#'   >,Period,Period >,Period,numeric >,numeric,Period >=,Duration,Period
-#'   >=,Period,Duration >=,Period,Period >=,Period,numeric >=,numeric,Period
+#'   -,ANY,Period-method -,Period,Period-method %%,Period,Duration-method
```
DavisVaughan, line 116: This is an annoying diff due to the way this documentation block is set up, but all I've done here is add `-,Period,Period-method` and then used RStudio to reformat to fit within the line length limits.

vspinu:
Thanks @DavisVaughan do you still want to fix anything? I would like to start a release process today.

DavisVaughan:
@vspinu nope, should be good to release!

PR #1010, opened 2021-12-31
Remember to use 'as_datetime' in 'round_date'

### LINKED ISSUE #1008: `round_date` compares days to seconds, leading to incorrect results
It seems that https://github.com/tidyverse/lubridate/commit/f4a4e426588729c0b781613ddc0fa688aaf29b68 has introduced a bug causing `round_date(x = <Date>, unit = <"2 days" or longer>)` to return rounded Dates that are either (a) wrong or (b) right, but only by coincidence. See my [answer](https://stackoverflow.com/questions/70222688/round-date-function-returns-floor-date-instead-of-rounded-date/70224473#70224473) to a recent SO question, which explains the bug.

To emphasize the extent of the regression, here is a transparent example of `round_date` failing:

``` r
library("lubridate")
#> 
#> 
R.version.string
#> [1] "R version 4.1.2 (2021-11-01)"
packageVersion("lubridate")
#> [1] '1.8.0'
round_date(seq(as.Date("2021-01-01"), as.Date("2021-01-31"), by = 1), unit = "month")
#>  [1] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#>  [6] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#> [11] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#> [16] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#> [21] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#> [26] "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01" "2021-01-01"
#> [31] "2021-01-01"
```

where it is understood that the second half of the output should contain `"2021-02-01"`...

jaganmn:
Fixes #1008

### DIFF
--- a/R/round.r
+++ b/R/round.r
@@ -148,7 +148,7 @@ round_date <- function(x, unit = "second", week_start = getOption("lubridate.wee
       round.POSIXt(as_datetime(x, tz = tz(x)), units = lub2base_units[[basic_unit]])
     } else {
       above <- unclass(as.POSIXct(ceiling_date(x, unit = unit, week_start = week_start)))
-      mid <- unclass(x)
+      mid <- unclass(as_datetime(x, tz = tz(x)))
       below <- unclass(as.POSIXct(floor_date(x, unit = unit, week_start = week_start)))
       wabove <- (above - mid) <= (mid - below)
       wabove <- !is.na(wabove) & wabove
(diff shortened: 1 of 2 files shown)


vspinu:
Thanks!

PR #1022, opened 2022-02-24
`use_z_format` argument added to `format_ISO8601()`

### LINKED ISSUE #941: Feature request: option to use "Z" tz label in format_ISO8601() output
This suggestion references @billdenney 's [comment](https://github.com/tidyverse/lubridate/issues/629#issuecomment-407961461) in #629, with the addition of a specific use case.

Bill wrote:
> I didn't do the additional formatting for adding Z instead of -0000 for UTC because that would have added a notable amount more code. I can do it, but the longer-term maintenance didn't seem like it was worthwhile. If you'd like Z, I can make that update.

### Example use case
The Twitter API v2 ([example](https://developer.twitter.com/en/docs/twitter-api/tweets/timelines/api-reference/get-users-id-tweets)) currently only accepts `YYYY-MM-DDTHH:mm:ssZ` format for the `start_time` and `end_time` query parameters. Passing in `YYYY-MM-DDTHH:mm:ss+0000` format results in a HTTP 400 Bad Request error. This makes it incompatible with `lubridate::format_ISO8601()`'s output.

``` r
lubridate::format_ISO8601(
  lubridate::make_datetime(2019),
  usetz = TRUE)
#> [1] "2019-01-01T00:00:00+0000"
```

It's not difficult for me to construct the "Z" format locally within my script and use that, but it would be great if this could be made part of `lubridate`'s capability. Something like:

``` r
# pseudocode
lubridate::format_ISO8601(
  lubridate::make_datetime(2019),
  use_z_format = TRUE)
#> [1] "2019-01-01T00:00:00Z"
```

This option obviously is only appropriate when the `tz` is UTC, or more realistically, where no timezone is stipulated. Would require some thought as to how it would work in cases when a date with a timezone is passed in. Just return an error, I suppose.

I'm happy to work on a PR to try to provide this feature, but I haven't looked into the code yet.

PS I'm adding the inverse feature suggestion on Twitter's API feedback forum - that they should accept all valid ISO8601 datetime formats, but we don't have much control over what they do!

billdenney:
Fix #941 

Adds a `use_z_format` argument to `format_ISO8601()` and does some code style reformatting within the same file.


vspinu:
Thanks, but this PR has a lot of white-space changes and it's dificult for me to judge what is actually changed. 

Regarding the name, would it be ok to use `z_format =` instead of `use_z_format`. We can also reuse `usetz = "Z"` argument. I think I would prefer the later option.

billdenney:
Sorry about the whitespace changes.  I tried to put them into the first commit because I saw that lines were getting pretty long.  But because some of the other code changes were already in place, the first commit does have a few other changes related to documentation.  If helpful, I can redo the PR without the whitespace changes.

Good call on the name.  I revised to `usetz="Z"`.

vspinu:
There are failures. Hadley is working on a new workflow, hopefully those are gone once done. Will merge this one once that works has been done.

januz:
When is this planned to be merged? Thanks!

vspinu:
@billdenney I made a few small changes and removed all the extraneous formatting which didn't match our house rules. 

Thanks!

billdenney:
Thanks for the effort to get it in!

PR #1023, opened 2022-02-24
Fix testing changes due to new vctrs outputs

billdenney:
An unrelated testing error popped up with committing #1022.  The error is due to format changes in the vctrs error messages.  This updates those tests.

There is another test (unrelated to #1022 as well) that is failing on my local system.  I'm not sure what the correct resolution is to that, so I will raise an issue for it.


hadley:
Fixed as part of #1029

PR #1029, opened 2022-03-03
Update to testthat 3e

hadley:
Nothing too surprising here except:

* I needed to add a few extra S4 methods to avoid messages about ambiguity that the 3e now lets through
* `all.equal()` automatically converted `POSIXlt` to `POSIXct`, which I now do manually


vspinu (approved):

hadley:
I'm going to assume the failures are due to a bad install and will get fixed after I merge the updated actions. If not, I'll investigate.

PR #1030, opened 2022-03-03
Restyle package

hadley:
Opening for the purposes of discussion — I didn't review it closely, but overall I feel like it's an improvement (which isn't surprising as I wrote the style guide that styler used here). @vspinu what do you think? If you're ok with it, I'd restyle after the changes in #1029.


vspinu (approved):
Looks good. Good idea. Over the years many styles were clashing here.

PR #1033, opened 2022-03-07
initial work setting week_start by string

### LINKED ISSUE #1017: FR: week_start= to accept "commonly recognized" weekday names for readability
When reading code,

```
floor_date(x, "week", week_start = 1)
```

is a bit hard to read. Is this week 0-or 1-based? Is it based on Sunday or Monday?

This is easily answered by consulting `?floor_date`, but when reading code / doing code reviews, it may not be trivial to do so (e.g. I don't have R open).

Therefore I think it makes sense for the following to "just work":

```
floor_date(x, "week", week_start = "Monday")
```

The only issue is this doesn't align perfectly with `format(x, "%A")` due to locale dependency. We could also do something like compare the input to `format(x, "%A")` at run-time (possibly with caching)... not sure it's worth it.

Happy to file a PR to this end.

MichaelChirico:
Closes #1017

Initial work; filing here to get high-level agreement before finishing tests & all. Also wasn't having any luck fidgeting with `LC_TIME` to try and get non-English week names to show up so we can test that too, any ideas?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,7 +54,7 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.2
+RoxygenNote: 7.2.1
 SystemRequirements: C++11, A system with zoneinfo data (e.g.
     /usr/share/zoneinfo). On Windows the zoneinfo included with R is used.
 Collate:
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@ Version 1.8.0.9000
 
 ### NEW FEATURES
 
+* [#1017](https://github.com/tidyverse/lubridate/issues/1017) `week_start` argument in all lubridate functions now accepts full and abbreviated names of the days of the week.
+* The assignment value `wday<-` can be a string either in English or as provided by the current locale.
 * Date rounding functions accept a date-time `unit` argument for rounding to a vector of date-times.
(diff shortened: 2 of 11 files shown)


MichaelChirico:
> The abbreviated names should also be supported.

this is done with `pmatch()`:

```
for (abbr in lubridate:::.get_locale_regs()$wday_names$abr) writeLines(format(round_date(ct, "week", week_start = abbr), "%a/%A"))
Sun/Sunday
Mon/Monday
Tue/Tuesday
Wed/Wednesday
Thu/Thursday
Fri/Friday
Sat/Saturday
```

My thinking was it's nice that it also works for less-canonical but also-accepted abbreviations:

```
for (abbr in c("Tues", "Thur", "Thurs")) writeLines(format(round_date(ct, "week", week_start = abbr), "%a/%A"))
Tue/Tuesday
Thu/Thursday
Thu/Thursday
```

> maybe a better option would be to construct a long vector possible matches and then take a modulo 7 of the match.

Nice idea! I wonder if maybe it's too clever though. e.g. it assumes the week always has 7 days (a quick search shows that there a very small number of places not using a 7-day week... I'm not sure if they have any support in `LC_TIME`, however).

Happy to switch to that if you'd like & are OK losing the extended matches from `pmatch()`.

> Regarding locale tests, we currently don't do that but would be good to have, but that would require adding extra languages to the CI. The plan is to move all locale suport directly into the package (like readr does it), but no time so far.

OK, noted, I'll not bother trying to add tests for this quite yet (though please do test manually if you're using a non-English locale)

vspinu:
Completed! Thanks for starting the work!

PR #1068, opened 2022-09-28
Fix R-devel CRAN checks: use format(), not as.character() for time/date

mmaechler:
Use  format()  not as.character()   for  POSIX[cl]t  and Date   --- will pass CRAN checks with R-devel


vspinu:
Thanks @mmaechler !

Issue #1070, opened 2022-10-05
Period addition and `update()` don't apply tidyverse recycling rules


DavisVaughan:
``` r
library(lubridate)

# should be size 0 result
as.Date("2019-01-01") + months(integer())
#> [1] "2019-01-01"

# should not be allowed!
as.Date(c("2019-01-01", "2019-01-02")) + months(1:3)
#> Warning in mt$mon + mos: longer object length is not a multiple of shorter
#> object length
#> Warning in mt$mday != ndays: longer object length is not a multiple of shorter
#> object length
#> Warning in mday(lt) + per@day: longer object length is not a multiple of shorter
#> object length
#> [1] "2019-02-01" "2019-03-02" "2019-04-01"

# like this:
as.Date(c("2019-01-01", "2019-01-02")) + days(1:3)
#> Warning in mday(lt) + per@day: longer object length is not a multiple of shorter
#> object length
#> Error: C_update_dt: length of dt vector must be 1 or match the length of updating vectors
```

``` r
library(lubridate)

# update() tries to follow tidyverse recycling rules
update(as.Date(c("2019-01-01", "2019-01-02")), days = 1:3)
#> Error: C_update_dt: length of dt vector must be 1 or match the length of updating vectors

# but it isn't quite right either (should be an error)
update(as.Date(c("2019-01-01", "2019-01-02")), days = integer())
#> [1] "2019-01-01" "2019-01-02"

# (should be size 0 result b/c we take the common size)
update(as.Date(c("2019-01-01")), days = integer())
#> [1] "2019-01-01"

# proof that we are generally taking the common size:
update(as.Date(c("2019-01-01")), days = 1:3)
#> [1] "2019-01-01" "2019-01-02" "2019-01-03"
```

<sup>Created on 2022-10-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

vspinu:
```R
# but it isn't quite right either (should be an error)
update(as.Date(c("2019-01-01", "2019-01-02")), days = integer())
#> [1] "2019-01-01" "2019-01-02"

# (should be size 0 result b/c we take the common size)
update(as.Date(c("2019-01-01")), days = integer())
#> [1] "2019-01-01"
```
Why should first one be an error and the second one not?

vspinu:
Fixed in devel. Though the error messages should be improved in the future.

```R
> update(as.Date(c("2019-01-01", "2019-01-02")), days = 1:3)
[...]

Issue #1069, opened 2022-10-02
floor_date() does not preserve NA on R devel


nealrichardson:
Ran into this very fine edge case in https://github.com/apache/arrow/pull/14282#issuecomment-1264031491. Due to a recent change in R devel (my guess is https://github.com/wch/r-source/commit/4f70ce0d79eeda7464cf97448e515275cbef754b), `floor_date()` can erroneously return 1970-01-01 for an NA Date input.

```
# Current R:
> lubridate::floor_date(as.Date(NA), "1 day")
[1] NA

# R at commit 82963
> lubridate::floor_date(as.Date(NA), "1 day")
[1] "1970-01-01"
```

I dug a bit and saw that it has to do with `lubridate:::floor_multi_unit1()`: on NA input, it returns `NA_real_` instead instead of the `integer` type that belongs in all of the fields of a `POSIXlt` object. This used to be fine:

```
# Current R
> as.Date(structure(list(sec =  0L, min =  0L, hour =  0L, mday = NA_real_, 
+     mon = NA_integer_, year = NA_integer_, wday = NA_integer_, 
+     yday = NA_integer_, isdst = -1L), class = c("POSIXlt", "POSIXt"
+ ), tzone = "UTC"))
[1] NA
```

But in R devel, it no longer does what is expected:
```
> as.Date(structure(list(sec =  0L, min =  0L, hour =  0L, mday = NA_real_, 
+     mon = NA_integer_, year = NA_integer_, wday = NA_integer_, 
+     yday = NA_integer_, isdst = -1L), class = c("POSIXlt", "POSIXt"
+ ), tzone = "UTC"))
[1] "1970-01-01"
```

though if `mday` were integer and not real, you get the expected `NA` result:

```
> as.Date(structure(list(sec =  0L, min =  0L, hour =  0L, mday = NA_integer_, 
+     mon = NA_integer_, year = NA_integer_, wday = NA_integer_, 
+     yday = NA_integer_, isdst = -1L), class = c("POSIXlt", "POSIXt"
+ ), tzone = "UTC"))
[1] NA
```

Arguably, R should handle this correctly. Happy to report this upstream if you think that's best. Though one could also argue that sticking something that's not an `integer` in a POSIXlt is breaking the data contract. (That said, I didn't see anywhere in the R docs that said all fields had to be integers, and to support fractional seconds, it would have to accept numeric at least in `secs`.) 

[...]

PR #1071, opened 2022-10-05
Recycle all POSIXlt fields when doing period month/year addition

DavisVaughan:
Part of https://github.com/tidyverse/lubridate/issues/1069, specifically my comment in https://github.com/tidyverse/lubridate/issues/1069#issuecomment-1268892740

So now we get this in the current state of R-devel again

``` r
as.Date("2019-01-30") + months(0:3)
#> [1] "2019-01-30" NA           "2019-03-30" "2019-04-30"
```

I've checked that this fixes slider too.

I'm convinced that this was an R-devel bug and I've sent them a message here about it https://stat.ethz.ch/pipermail/r-devel/2022-October/082066.html. Nevertheless, I feel like we could be more robust about recycling the fields of the POSIXlt objects where possible to avoid these issues altogether. Relying on base R's hidden recycling of the POSIXlt fields seems dangerous, and I'm not sure it is really documented well.

We also actually use base R recycling rules (accidentally?) here due to the `+` call. I've opened https://github.com/tidyverse/lubridate/issues/1070 to track a move towards tidyverse recycling rules, but for now I have tried to keep the base R recycling rules intact.

This patch doesn't actually fix the `floor_date()` issue in #1069. But it is possible the fixing of the r-devel bug itself will fix that issue. I am not sure yet, but will look at that in another PR. It probably implies we are partially recycling POSIXlt objects somewhere else, so it would be good to track it down either way.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@ Version 1.8.0.9000
 
 ### BUG FIXES
 
+* [#1069](https://github.com/tidyverse/lubridate/issues/1069) Internal code handling the addition of period months and years no longer generates partially recycled POSIXlt objects.
 * Fix rounding of POSIXlt objects
 * [#1007](https://github.com/tidyverse/lubridate/issues/1007) Internal lubridate formats are no longer propagated to stamp formater.
 * `train` argument in `parse_date_time` now takes effect. It was previously ignored.
--- a/R/ops-addition.r
+++ b/R/ops-addition.r
@@ -89,8 +89,9 @@ add_months <- function(mt, mos) {
   }
 
   if (anyNA(new_mon)) {
-    # Ensure missing value assignment propagates to all fields
-    mt[is.na(new_mon)] <- NA
+    # Ensure missing value assignment propagates to all fields.
+    # Can't just assign `NA` https://stat.ethz.ch/pipermail/r-devel/2022-October/082080.html.
+    mt[is.na(new_mon)] <- as.POSIXlt(NA, tz = tz(mt))
   }
 
   mt$mon <- new_mon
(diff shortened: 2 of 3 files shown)


DavisVaughan:
oldrel-4 is failing because of `timeDate: Needs R >= 3.6.0 `

vspinu:
The link to the report does not work for me. An alternative https://hypatia.math.ethz.ch/pipermail/r-devel/2022-October/082080.html

The patch looks good to me but I would rather see this fixed in R-devel instead. Adding so much code to fix a broken NA assignment doesn't seem right to me. Many others are likely affected, so should be fixed upstream. What do you think?

DavisVaughan:
It was actually _two_ problems in r-devel.

The PR was motivated by this one (not sure how I messed up the link before):
https://stat.ethz.ch/pipermail/r-devel/2022-October/082075.html

And while working on it I found the other one you pointed to:
https://stat.ethz.ch/pipermail/r-devel/2022-October/082080.html

We seem to rely on POSIXlt partial recycling or `NA` propagation in two places (that I am aware of). Here, and in `floor_date()` (which is what actually caused #1069) 
https://github.com/tidyverse/lubridate/blob/be024010ededa0d09cffe6e323777e3ab266f419/R/round.r#L381-L395

Ideally I think we wouldn't rely on this at all. It technically isn't a documented feature of POSIXlt, so I'd like to try and remove our dependence on it.

vspinu:
@DavisVaughan I am still not sure about this PR. Judging from the threads there is was a considerable effort from Martin to address the issues. And it looks like they have been already addressed in devel. No?

If we were to deal with this on our side then it would require adding an internal helper (normalize_POSIXlt or something, or is it the same as balancePOSIXlt that Martin already implemented in devel?) and then hunting all the occurrences of implicit recycling in our code in order to recycle.  Thus this PR is somewhat half way through in that regard.

Shall we merge all the tests that you have created and then keep thinking of how to proceed further?

vspinu:
I am a bit confused. I thought this is only R devel issues, but I see the same in R4.1 as well. So we do need to recycle on our side indeed :(.

vspinu:
@DavisVaughan I have merged this except the last commit (given that it is already fixed in devel). Thanks!

PR #1078, opened 2022-10-31
Move lubridate on top of timechane

vspinu:
- CCTZ and cpp11 dependency is gone


vspinu:
@DavisVaughan @hadley I have in plan to merge this tomorrow as I am pressed by a CRAN deadline on a deadline now. 

This change fixes a number of critical issues, simplifies lubridate's code and streamlines DST and leap logic throughout the package. So if it really comes to porting lubridate on top of clock some day it would be much easier to do it from this new state.

hadley:
If you need to get this out for a CRAN release, go for it, but I'd like us to find some time to chat about lubridate in the future. In particular, for these changes needed by CRAN fixes you should feel free to ping me or Davis for help because we normally have to fix them for many packages so it's normally pretty easy for us (plus we're paid to do this full time).

PR #1081, opened 2022-10-31
Place names on the `.Data` column of the proxy

### LINKED ISSUE #1072: Period and Interval vctrs proxy methods need to tweak handling of names
The fact that `rcrd_names` is an _optional_ column for the `vec_proxy()` method poses a problem for combination functions

``` r
x <- `names<-`(period(1), "x")
y <- period(2)

vctrs::vec_c(x, y)
#> Error in `vctrs::vec_c()`:
#> ! Can't assign 7 columns to df of length 6.
#> ℹ In file 'slice-assign.c' at line 341.
#> ℹ This is an internal error that was detected in the vctrs package.
#>   Please report it at <https://github.com/r-lib/vctrs/issues> with a reprex (<https://https://tidyverse.org/help/>) and the full backtrace.

#> Backtrace:
#>     ▆
#>  1. ├─vctrs::vec_c(x, y)
#>  2. └─rlang:::stop_internal_c_lib(...)
#>  3.   └─rlang::abort(message, call = call, .internal = TRUE, .frame = frame)

x <- `names<-`(interval(.Date(0), .Date(2)), "x")
y <- interval(.Date(0), .Date(3))

vctrs::vec_c(x, y)
#> Error in `vctrs::vec_c()`:
#> ! Can't assign 3 columns to df of length 2.
#> ℹ In file 'slice-assign.c' at line 341.
#> ℹ This is an internal error that was detected in the vctrs package.
#>   Please report it at <https://github.com/r-lib/vctrs/issues> with a reprex (<https://https://tidyverse.org/help/>) and the full backtrace.

#> Backtrace:
#>     ▆
#>  1. ├─vctrs::vec_c(x, y)
#>  2. └─rlang:::stop_internal_c_lib(...)
#>  3.   └─rlang::abort(message, call = call, .internal = TRUE, .frame = frame)
```

I think the solution is actually pretty easy. Since these are S4 objects it looks like the names are already being put on `@.Data`, so we may as well leave them there when we promote to a data frame to generate the proxy. The names will get sliced accordingly and then will be restored automatically. This is similar to how it works in clock now.

DavisVaughan:
Closes #1072 

We now store the names on the column that corresponds to the `@.Data` slot of the original object. For Period this is `$second` and for Interval this is `$span`. This fixes the problem of #1072 because it means that we never have a mismatch in the number of columns that are on the proxy.

A few extra niceties here:
- No need for compare/equal proxy methods anymore, we just carry around the names unconditionally, which doesn't hurt anything
- No need for the complicated recursive proxying done in the Interval methods, because in vctrs 0.5.0 if a data frame is returned as the proxy then it will automatically proxy the columns recursively too (CC @lionel- this is a nice side effect of that change we made)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,7 +39,7 @@ Suggests:
     knitr,
     rmarkdown,
     testthat (>= 2.1.0),
-    vctrs (>= 0.4.1)
+    vctrs (>= 0.5.0)
 Enhances:
     chron,
     data.table,
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,7 @@ Version 1.8.0.9000
 
 ### BUG FIXES
 
+* [#1072](https://github.com/tidyverse/lubridate/issues/1072) Names are now handled correctly when combining multiple Period or Interval objects.
 * [#1003](https://github.com/tidyverse/lubridate/issues/1003) Correctly handle r and R formats in locales which have no p format
 * [#1074](https://github.com/tidyverse/lubridate/issues/1074) Fix concatination of named Period, Interval and Duration vectors.
 * [#1044](https://github.com/tidyverse/lubridate/issues/1044) POSIXlt results returned by `fast_strptime()` and `parse_date_time2()` now have a recycled `isdst` field.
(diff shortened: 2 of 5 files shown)


vspinu (approved):
Looks good to me, but I am really not dwelling into all those vctrs nuances. Please feel free to merge if you think its ready to go.

Issue #1082, opened 2022-11-02
Difficult to define custom "tz<-" method


trevorld:
* Unlike most of `{lubridate}`'s accessors it is difficult to define a custom "tz<-" method
* `tz<-()` is not a generic method.  It calls `force_tz()` which is also not a generic method.
* It would be easier if `force_tz()` was a generic S3 (or S4) method

PR #1083, opened 2022-11-02
feat: Generic 'force_tz()'

trevorld:
* Makes `force_tz()` a generic S3 method
* Current `force_tz()` becomes new default method `force_tz.default()`
* This would allow supporting `tz<-()` for new datetime classes (#1082).


vspinu:
Can you add with_tz as well then?

trevorld:
> Can you add with_tz as well then?

Okay, I've added a generic S3 method `with_tz()` as well.  The previous `with_tz()` is now the default method `with_tz.default()`.

vspinu:
Thanks!

PR #1094, opened 2022-11-10
Move timechange from Depends to Imports

### LINKED ISSUE #1090: Can timechange be in Imports, not Depends?
I don't think there is any reason it should be in Depends, right? It forces lubridate to show this message when it is loaded

```r
#> Loading required package: timechange
```

We generally prefer putting packages in Imports unless there is a _really_ compelling reason not to

DavisVaughan:
Closes #1090

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,9 @@
-# lubridate (development version)
+Version 1.9.0.9000
+=============
+
+### BUG FIXES
+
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,4 +1,9 @@
-# lubridate (development version)
+Version 1.9.0.9000
+=============
```
DavisVaughan, line 2: This also didn't seem to match the way the NEWS typically looked, so I tweaked it

vspinu:
Thanks!

Issue #1100, opened 2022-12-08
Confusing documentation vs default value for week_start in wday()


FlorenceMounier:
Hello,

My problem is somehow related to #1025 but for a different reason and with different propositions to solve the problem.

I find the choice of default value for `week_start` really confusing in `wday()`, especially for beginners (see "Details" section at the end of this issue if necessary).

However, I understand that this is an historical choice and that changing it would have many consequences for users. Having a more explicit documentation (1), or adding an example(2), would probably be the best options.

1.  Rewording

`week_start`: day on which week starts following ISO conventions - 1 means Monday, 7 means Sunday (default). **_When label = FALSE, the number returned for this day of the week will be 1._** When label = TRUE, this day **_of the week_** will be the first level of the returned factor. You can set lubridate.week.start option to control this parameter globally.`

2. New example

```r
library(lubridate)

x <- ymd("2022-12-25")
wday(x)  # 1
wday(x, label = TRUE) 
#> [1] Sun
#> Levels: Sun < Mon < Tue < Wed < Thu < Fri < Sat
```

---
### Details
The documentation of `wday()` for the `week_start` argument mentions that `"following ISO convention [...] 7 means Sunday (default)"`. As a consequence, we could expect the result of `wday()` to be equal to `7` for a `Sunday`, by default.

However, `wday()` returns `1` for a Sunday as the default value for `week_start` is `7`, thus `Sunday` is the first day of the week.

The documentation, and the use of `wday()`, would be clearer if the default value for `week_start` was consistent with the ISO conventions used and mentionned (i.e. 1 means Monday, thus `week_start` would be 1 by default).
Indeed, it would be much easier to explain to beginners.

GegznaV:
I agree that consistency is essential especially when you have to teach. Every inconsistency consumes time to explain and it becomes more difficult to remember. I would vote for consistency, clear documentation and **for following ISO standards**.  But I know (as a user and as a package maintainer) that it is inconvenient to change the default behaviour of functions.  But this sentence in the documentation:

<html>
<body>


`week_start` | day on which week starts following ISO conventions - 1 means Monday, 7 means Sunday (default).
-- | --



</body>
</html>

[...]

PR #1101, opened 2022-12-10
Improve examples in `wday()`

GegznaV:
A better illustration of the relationship between weekday name in words and weekday number as well as between weekday number and the value of argument `week_start`. 
Related to #1100

### DIFF
--- a/R/accessors-day.r
+++ b/R/accessors-day.r
@@ -29,6 +29,7 @@ NULL
 #' @examples
 #' x <- as.Date("2009-09-02")
 #' wday(x) # 4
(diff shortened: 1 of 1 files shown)


GegznaV:
Should I also update `news.md`? If yes, should I create a new section `### Improvements` or add a bullet in the bug fix section?

vspinu:
Merged offline (e7b88608) with your other PR. Added a few more examples and brought even more clarity in `label=TRUE` case. 

Thanks!

GegznaV:
Thank you for merging.

PR #1103, opened 2022-12-15
Use explicit integers in `year()` and `month()`

malcolmbarrett:
An astute student pointed out to use that `day()`, `year()` and `month()` are inconsistent in their return types:

``` r
library(lubridate)
x <- ymd("2012-03-26")
is.integer(month(x))
#> [1] FALSE
is.integer(day(x))
#> [1] TRUE
is.integer(year(x))
#> [1] FALSE
```

<sup>Created on 2022-12-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This seems like an unintentional conversion due to addition. This PR uses explicit integer addition to have `month()` and `year()` return integers

### DIFF
--- a/R/accessors-month.r
+++ b/R/accessors-month.r
@@ -35,7 +35,7 @@ month <- function(x, label = FALSE, abbr = TRUE, locale = Sys.getlocale("LC_TIME
 
 #' @export
 month.default <- function(x, label = FALSE, abbr = TRUE, locale = Sys.getlocale("LC_TIME")) {
-  month(as.POSIXlt(x, tz = tz(x))$mon + 1, label, abbr, locale = locale)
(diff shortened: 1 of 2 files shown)


vspinu:
This has been [attempted before](https://github.com/tidyverse/lubridate/pull/814#issuecomment-602043913) The consequences are not as benign as they might appear. 

Let's try to give it another try, but not for the forthcoming release (end of this month) but for the next one. leaving this open till then.

malcolmbarrett:
Sounds good. Happy to help with revdep fallout when the time comes. Just let me know

krlmlr:
I went ahead and ran the revdepchecks: https://github.com/krlmlr/lubridate/blob/make-integers/revdep/problems.md . Only two packages affected now. Should we give it a try?

PR #1118, opened 2023-05-09
docs: use period() in period's examples

eitsupi:
Given the context, `period()` seems appropriate, not `duration()`.


vspinu:
Thanks!

Issue #1117, opened 2023-05-02
Search conda path for timezone database


drhagen:
When installing `lubridate` into a conda environment, lubridate does not see the timezone data stored at `${CONDA_PREFIX}/share/zoneinfo`. I see the typical warning message like this and none of the timezones work:

```
Warning message in system("timedatectl", intern = TRUE):
“running command 'timedatectl' had status 1”
Warning message:
“Failed to locate timezone database”
```

I know this is kind of a long shot since Tidyverse is not a conda project, but would it be reasonable to add `${CONDA_PREFIX}/share/zoneinfo` to the [list of default paths](https://github.com/tidyverse/lubridate/blob/main/R/tzdir.R#L31-L53) that lubridate searches.

I know I can set `TZDIR`, but that is [kind of impossible](https://stackoverflow.com/questions/76157408/conda-environment-variable-referencing-another-variable) when creating a conda environment file for someone else.

vspinu:
I think this addition would not harm but I am not a frequent user of conda even on Python side. So cannot judge. Is Conda still going strong? I would not want to add support for a infrastructure project that gonna die in a few years anyhow. Especially that it didn't seem to have caught up with R folks so far.

@hadley any opinion on this one?

hadley:
I have little to add hear, except that anecdotally using condas with R seems to be a great struggle. OTOH this seems like a relatively small amount of work, and it would ease one papercut that faces conda + R users.

drhagen:
> Is Conda still going strong? I would not want to add support for a infrastructure project that gonna die in a few years anyhow.

I would say that conda is more popular than it has ever been. It is an essential component of scientific Python. This has been largely driven by Jupyter, which adds Javascript to the mix of Python and C needed to set up a scientific Python environment. Conda can easily install packages from different programming languages together, so it is great for this.

If you don't want to take my word for it, [Conda](https://insights.stackoverflow.com/trends?tags=conda) is tagged in about 0.1% of all Stack Overflow posts and still growing. That is pretty high for something on SO that is not a programming language.

drhagen:
Yes, I will make a PR for this.

PR #1119, opened 2023-05-12
Use CONDA_PREFIX for timezone database if available

drhagen:
As discussed in #1117, this allows lubridate to use the timezone database in a conda environment if R is currently being executed in one. The conda environment is detected and located using the `CONDA_PREFIX` environment variable.

The behavior of lubridate was tested before and after this change. Indeed, the timezone information that is installed by the standard `tzdata` conda package at `${CONDA_PREFIX}/share/zoneinfo` is detected and used.

Please check the logic around the handling of Windows. Because conda environment can exist on Windows (and are actually a very nice way to set up a scientific programming environment there), I reorganized the way `tzdir_find` works to avoid too much nesting.

### DIFF
--- a/R/tzdir.R
+++ b/R/tzdir.R
@@ -29,21 +29,29 @@ tzdir_get <- function() {
 
 ## adapted from Syz.timezone and OlsonNames function
 tzdir_find <- function() {
-  if (.Platform$OS.type == "windows") {
-    return(tzdir_internal())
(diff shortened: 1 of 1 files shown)


vspinu:
Thanks! Looks good. The windows treatment should be fine. Those paths don't make sense on windows indeed.

PR #1136, opened 2023-09-28
Make interval equality operator compare start and end of intervals

### LINKED ISSUE #1135: Equal to ==  operator doesn't work with lubridate intervals
If you attempt to use an `==` check with lubridate intervals, it always returns `TRUE`.
This does not happen when using dates.

```r
lubridate::interval(start='2020-01-01', end='2020-01-02') == lubridate::interval(start='2020-01-03', end='2020-01-04') 
#> [1] TRUE
> lubridate::as_date('2020-01-01') == lubridate::as_date('2020-01-02')
#> [1] FALSE
```

jakemanger:
An approach to Fix #1135 
This changes the default behaviour from comparing the interval duration to comparing the start and end of the interval, which I believe is more intuitive (it has slipped me up before). However, would this require a warning to notify users that the default behaviour has changed?


vspinu:
Hi @jakemanger, there is a failure in pkgdown. I think you need to declare it in the index. 

More importantly, Should we add inequalities as well, for full partial order implementation?

Also could you please test for vectorized operations, not just single-tons, and add a new-features news entry? 

Thanks!!!

vspinu:
Hey @jakemanger sorry for the death ear for a year :( I was out of maintenance for a while. Indeed, the interval objects are actually vectorized objects, so the equality should be done element wise. The vctrz part is also important, but probably will be fixed once the operation is vectorized. 

And fineally, it's a breaking change indeed. A carefull revdep has to be run before merging this one. If it impacts a lot of dependencies then we are in trouble.

PR #1138, opened 2023-10-02
timechange dependency is 0.2.0

MichaelChirico:
Alternatively, since the newer version is really only required in tests, we can `skip_if_not_installed("timechange, "0.2.0")` here:

https://github.com/tidyverse/lubridate/blob/e54b612c0672c57bb9cea12dd2482b0d59b3aa64/tests/testthat/test-round.R#L655


vspinu:
Thanks!

Issue #1171, opened 2024-11-02
When `lubridate::guess_formats()` is given a vector of length greater than 1, return a matrix


guslipkin:
When `lubridate::guess_formats()` is given a vector of length greater than 1, it returns a vector that has had some form of vector recycling performed. This is less than useful because there is no easy way to associate a given input with a given output. However, the `print_matches` argument causes a very nice matrix to print out where each row corresponds with an item in the input vector and each column with a value in the `orders` argument. However, because the matrix is printed, there is no easy way to capture it. When `x` is a vector of length greater than one, that matrix should be returned and not the vector.

``` r
# the current behavior
lubridate::guess_formats(
  x = c('12-10-2024', '13-10-2024'), 
  orders = c('mdY', 'dmY'),
)
#>        OmdY        dOmY        dOmY         mdY         dmY         dmY 
#> "%Om-%d-%Y" "%d-%Om-%Y" "%d-%Om-%Y"  "%m-%d-%Y"  "%d-%m-%Y"  "%d-%m-%Y"
# the proposed behavior
lubridate::guess_formats(
  x = c('12-10-2024', '13-10-2024'), 
  orders = c('mdY', 'dmY')
)
#>      x             OmdY        dOmY        mdY        dmY       
#> [1,] "12-10-2024" "%Om-%d-%Y" "%d-%Om-%Y" "%m-%d-%Y" "%d-%m-%Y"
#> [2,] "13-10-2024" ""          "%d-%Om-%Y" ""         "%d-%m-%Y"
```

<sup>Created on 2024-11-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1172, opened 2024-11-02
If length(x) > 1, return a matrix

guslipkin:
This sketches a possible solution to #1171


guslipkin:
This appears to break many things as the long vector is apparently now expected and relied upon behavior. Perhaps an option then to return the matrix instead?

vspinu:
Yes, an option to return a matrix is a much better idea. Could get rid of the print method. Those are for debugging purposes anyhow.

PR #1175, opened 2024-12-13
Created set_term function and clean_class_days function

hersheyk:
Creates a term data frame, allowing personalized academic terms to be created.

 cc: @kbodwin to review


vspinu:
Sorry, I don't understand what `term` means.  These are the questions what I would like to know the answer before diving further:

- What is the "academic term"
- What are the concrete usecases?
- Why cannot it be directly set with existing tooling in lubridate?
- Why should it live in lubridate?

Issue #902, opened 2020-06-20
More vctrs support (min, max, seq) for Period


earowang:
Since `Period` class can be sorted now, shall we expect `vec_math()` to return the class itself?

``` r
p <- lubridate::period(year = 1:5)
p
#> [1] "1y 0m 0d 0H 0M 0S" "2y 0m 0d 0H 0M 0S" "3y 0m 0d 0H 0M 0S"
#> [4] "4y 0m 0d 0H 0M 0S" "5y 0m 0d 0H 0M 0S"
min(p)
#> [1] 0
max(p)
#> [1] 0
range(p)
#> [1] 0 1
```

**Feature request**: It would be useful to add the `seq()` method for `Period`, in addition to its arithmetics.

```r
p + lubridate::period(year = 1)
#> [1] "2y 0m 0d 0H 0M 0S" "3y 0m 0d 0H 0M 0S" "4y 0m 0d 0H 0M 0S"
#> [4] "5y 0m 0d 0H 0M 0S" "6y 0m 0d 0H 0M 0S"
seq(p[5], by = "1 year", length.out = 1)
#> Error in (0L:(length.out - 1L)) * by: non-numeric argument to binary operator
```

<sup>Created on 2020-06-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Happy to send PRs, if needed.

vspinu:
@DavisVaughan any thoughts?

DavisVaughan:
As of right now, `vec_math()` isn't being called for lubridate objects. I don't _think_ this currently has much to do with vctrs? I would imagine it is falling through to the default `min()` implementation unless lubridate has added one specifically for period objects.

It seems like a useful thing to add though

vspinu:
Related #918

moodymudskipper:
To add to the list:
*  `for(p in periods) {}` loops on the seconds
*  `ifelse(TRUE, period1, period2)` returns the seconds (`dplyr::if_else()` works fine).
* `period_1 | period_2` and `period_1 || period_2` are `FALSE` if seconds are zero, despite having non zero component
* `if (period_1)` similarly will take `period_1` as `FALSE` if it has zero seconds, even if it has other components

If the type was built on top of NA rather than 0 (and we'd have a separate seconds component) we would probably spot the issue more easily due to NA propagation, now the behavior is very dangerous in my opinion.

[...]

PR #1177, opened 2025-01-16
`sum()` method for `Duration` class

d-morrison:
re: #902

I'm trying to preserve the `Duration` class through `sum()`; for example:

Before: 

``` r
x = lubridate::duration(1:2)
sum(x)
#> [1] 3
```

After:

``` r
x = lubridate::duration(1:2)
sum(x)
#> [1] "3s"
class(x)
#> [1] "Duration"
#> attr(,"package")
#> [1] "lubridate"
```

<sup>Created on 2025-01-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Not sure if this will be helpful, but figured it was worth trying; glad to keep working on a better solution based on your feedback.
I used `purrr::reduce()`, which I'm guessing introduces an unwanted dependency.
I tried `base::Reduce()` but it doesn't preserves the `Duration` class unfortunately.


vspinu:
This is really nice addition but I would prefer to avoid adding extra dependency just for this change. It's straightforward to do in plain R.

PR #1180, opened 2025-02-06
Use R for Data Science 2nd Edition

jack-davison:
Noticed that the lubridate REAMDE & pkgdown points to https://r4ds.had.co.nz/dates-and-times.html, which itself then points to https://r4ds.hadley.nz/datetimes.html. I've gone in and changed both to just go straight to https://r4ds.hadley.nz/datetimes.html, and re-rendered the README, to save someone a job.


vspinu:
Thanks!

PR #1185, opened 2025-03-27
Update round.r, unit = "second"

erikerhardt:
Change unit = "seconds" to "second" to be consistent with other documentation


vspinu:
Thanks for the PR. The checks are failing. Could you run roxygen to update the docs?

